PR #13, opened 2018-06-14
wait for server to spawn before redirecting

### LINKED ISSUE #12: should wait for server to be ready before redirecting
If the notebook server takes a long time to spawn (e.g. because images have to be downloaded), JH will wait for 30 seconds and then redirect to a url like `.../hub/user/...` which fails with a 401 because it's not actually a valid URL (it still gets proxied to the notebooks service for some reason which queries gitlab for a non-existent project, hence the 401). 

The way to fix this is to wait until the server is ready before redirecting. It's unclear how to get this information.

moved from https://github.com/SwissDataScienceCenter/renku/issues/210 -- see https://github.com/SwissDataScienceCenter/renku/issues/210#issuecomment-394503121

rokroskar:
closes #12

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -21,6 +21,7 @@ import hashlib
 import json
 import os
 import string
+import time
 from urllib.parse import urljoin
 from functools import partial, wraps
 
@@ -133,10 +134,6 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
         headers=headers,
     )
 
-    # 2. redirect to launched server
-    if r.status_code not in {201, 202, 400}:
-        abort(r.status_code)
-
     notebook_url = urljoin(
         os.environ.get('JUPYTERHUB_BASE_URL'),
         'user/{user[name]}/{server_name}/'.format(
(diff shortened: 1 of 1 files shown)


src/notebooks_service.py
```diff
@@ -147,7 +144,32 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
...
 
-    return redirect(notebook_url)
+    # 2. redirect to the server when ready
+    if r.status_code == 201:
+        # server is already running
+        return redirect(notebook_url)
+
+    elif r.status_code == 202:
+        # server is spawning - wait a max of 900 seconds = 15 minutes
+        for i in range(450):
```
leafty: Is there a way to avoid magic numbers?

leafty: Is there a way to avoid magic numbers?

rokroskar: Sure - do you have a suggestion? I'm not sure what an appropriate time to wait would be exactly.

leafty: What you suggested is fine, it's just that linking 15 minutes to 450*2 happens multiple lines down the comment.
Maybe something like:
```python
timeout = 900 # 900 seconds = 15 minutes
sleep_time = 2
for i in range(timeout / sleep_time):
```

leafty: You can also use `datetime` objects with a while loop.

rokroskar: yep, I intended to parametrize this a bit better, I thought your concern was about the length of the timeout itself.

PR #15, opened 2018-06-15
Create UI for connecting to and stopping jupyter notebook servers.

### LINKED ISSUE #4: Create a notebooks landing page
The `<service-prefix>` endpoint right now just gives a JSON of the user object. It should give the authenticated user an overview of the running notebook servers and the means to stop them. 

One main issue to resolve is how to actually get the information about running servers per user. Right now this seems to be kept in the user object obtained through the `HubOAuthenticator` but this is cached via some (as of yet) to me obscure mechanism. Since the user info is cached, we don't get up-to-date info unless we reduce the caching timeout. There should be some way to query the hub directly, however. 

cc/ @ableuler @ciyer 

related to https://github.com/SwissDataScienceCenter/renku-ui/issues/151

ciyer:
Implements a basic UI for viewing, connecting to, and stopping running servers. Fix #4.


ciyer:
Two things are still outstanding:
- The UI should be built in the Dockerfile. This requires node.
- The URLs used by the UI should be derived from configuration. They are currently hard-coded.

leafty:
@ciyer @rokroskar I can have a look at building with node, with some multi-stage process.

rokroskar:
@leafty would be great so we can get this on staging -- I guess that https://github.com/SwissDataScienceCenter/renku-ui/blob/master/Dockerfile can be of use

src/notebooks_service.py
```diff
@@ -95,10 +97,28 @@ def health():
...
+            '{prefix}/users/{user[name]}'.format(
+                prefix=auth.api_url, user=user
+            ),
+            headers=headers
+        ).text
     )
+    return jsonify(info)
+
+
+# @authenticated
```
rokroskar: why is this commented out?

src/notebooks_service.py
```diff
@@ -95,10 +97,28 @@ def health():
...
+@app.route(urljoin(SERVICE_PREFIX, 'ui'), defaults={'path': ''})
+@app.route(urljoin(SERVICE_PREFIX, 'ui/'), defaults={'path': ''})
```
rokroskar: is this necessary? I didn't find that I needed to specifically include routes with trailing slashes

src/notebooks_service.py
```diff
@@ -95,10 +97,28 @@ def health():
...
+@app.route(urljoin(SERVICE_PREFIX, 'ui/<path:path>'), methods=['GET'])
+def ui(path):
+    """Route for serving a UI for managing running servers."""
+    if not path:
+        path = "index.html"
+    return send_from_directory('ui/build/', path)
```
leafty: The ui is a set of static assets, so they should be served with `url_for('static', filename=...)`

src/notebooks_service.py
ciyer: Sorry, this was a mistake. I was in a rush to get the code pushed to the server. I fixed this and turned the annotation on.

src/notebooks_service.py
ciyer: You are right, it is unnecessary. I have removed the route without the trailing slash, and it still works when I provide the URL without the trailing slash.

src/notebooks_service.py
```diff
@@ -95,10 +97,29 @@ def health():
...
+            '{prefix}/users/{user[name]}'.format(
+                prefix=auth.api_url, user=user
+            ),
+            headers=headers
+        ).text
     )
+    return jsonify(info)
+
+
+@authenticated
```
jirikuncar: why do have to be the static files accessible only for authenticated users?

src/notebooks_service.py
```diff
@@ -95,10 +97,29 @@ def health():
...
+@app.route(urljoin(SERVICE_PREFIX, 'ui/'), defaults={'path': ''})
+@app.route(urljoin(SERVICE_PREFIX, 'ui/<path:path>'), methods=['GET'])
+def ui(path):
+    """Route for serving a UI for managing running servers."""
+    if not path:
+        path = "index.html"
+    return send_from_directory('../static', filename=path)
```
jirikuncar: can we let flask to serve the static files directly? 

`app=Flask(__name__, static_url_path='ui', static_folder='static')`

jirikuncar: For production purposes we should serve it via nginx anyway.

src/notebooks_service.py
```diff
@@ -193,6 +214,25 @@ def stop_notebook(user, namespace, project, commit_sha):
     return app.response_class(r.content, status=r.status_code)
 
 
+@app.route(
+    urljoin(SERVICE_PREFIX, 'stop/<server_name>'),
```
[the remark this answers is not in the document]
ciyer: Yes, you are right, this is a better name.

src/notebooks_service.py
ciyer: Because the static files access API endpoints that require authentication. The UI code is simpler if it can assume that the user is already authenticated.

src/notebooks_service.py
ciyer: I did not know how to apply the authenticated decorator to serving static files when Flask is configured this way. If you can tell me how to do that, we can set up the server this way.

src/notebooks_service.py
jirikuncar: But what if we want to serve the files directly using nginx?

ciyer: My understanding of the requirements for #4 is that this solution should be minimal and simple. For Renku, we will create a more full-featured UI. That UI is already being served from nginx.

PR #17, opened 2018-06-18
redirect to existing server if already running

### LINKED ISSUE #16: clicking "launch" for an existing notebook yields "bad request"
If a server is already running, this is not handled properly.

rokroskar:
closes #16

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -147,6 +147,28 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
 
     headers = {auth.auth_header_name: 'token {0}'.format(auth.api_token)}
 
+    notebook_url = urljoin(
+        os.environ.get('JUPYTERHUB_BASE_URL'),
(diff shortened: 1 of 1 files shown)


src/notebooks_service.py
```diff
@@ -223,8 +235,7 @@ def stop_notebook(user, namespace, project, commit_sha):
 
 
 @app.route(
-    urljoin(SERVICE_PREFIX, 'servers/<server_name>'),
-    methods=['DELETE']
+    urljoin(SERVICE_PREFIX, 'servers/<server_name>'), methods=['DELETE']
```
leafty: keep old formating?

rokroskar: no, this is automatically formatted using the style from `.style.yapf`

rokroskar:
@leafty your concern regarding the 400 you get when trying to access a server that is being terminated is still a problem and I will open an issue for it.

PR #20, opened 2018-07-04
[Chart] Create a service account to watch pods

### LINKED ISSUE #19: create a k8s service account to watch pods
This is needed to better respond to problems during server spawn, e.g. when an image is not able to be pulled.

leafty:
This PR resolves #19.

The service account is functional, to test it when renku-notebooks is deployed:
```bash
$ kubectl -n renku run my-shell -i --tty --image ubuntu:16.04 bash
$ exit
$ kubectl -n renku edit deploy my-shell
<Put 'serviceAccountName: renku-notebooks' in the podspec>
$ kubectl -n renku attach my-shell-75d8cbc9f4-9sktv -c my-shell -i -t # it's a new pod
$ apt update && apt install -y curl
$ curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
$ chmod +x ./kubectl
$ mv ./kubectl /usr/local/bin/kubectl
$ kubectl get po # works
$ kubectl get cm # is denied
```

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -34,6 +34,8 @@ from flask import jsonify
 from flask import send_from_directory
 from jupyterhub.services.auth import HubOAuth
 
+from kubernetes import client, config
+
 SERVICE_PREFIX = os.environ.get('JUPYTERHUB_SERVICE_PREFIX', '/')
 """Service prefix is set by JupyterHub service spawner."""
 
@@ -47,6 +49,10 @@ auth = HubOAuth(
 """Wrap JupyterHub authentication service API."""
 
 app = Flask(__name__)
+
+config.load_incluster_config()
+with open('/var/run/secrets/kubernetes.io/serviceaccount/namespace', 'rt') as f:
+    kubernetes_namespace = f.read()
 
 
 def _server_name(namespace, project, commit_sha):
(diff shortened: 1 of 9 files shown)


rokroskar:
@leafty as a demo of this, could you add an endpoint that lists the status of the pods with the `heritage=jupyterhub` label?

src/notebooks_service.py
```diff
@@ -34,6 +34,8 @@
 from flask import send_from_directory
 from jupyterhub.services.auth import HubOAuth
 
+from kubernetes import client, config
```
rokroskar: there should be a try/except so that it can work with a docker-compose deployment also

src/notebooks_service.py
```diff
@@ -48,6 +50,10 @@
 
 app = Flask(__name__)
 
+config.load_incluster_config()
+with open('/var/run/secrets/kubernetes.io/serviceaccount/namespace', 'rt') as f:
```
rokroskar: needs a try/except or some other check that we're running on k8s

src/notebooks_service.py
```diff
@@ -252,6 +258,15 @@ def stop_server(user, server_name):
     return app.response_class(r.content, status=r.status_code)
 
 
+@app.route(
+    urljoin(SERVICE_PREFIX, 'pods'), methods=['GET']
+)
+@authenticated
+def list_pods(user):
+    v1 = client.CoreV1Api()
```
rokroskar: maybe return the docker equivalent if not running in kubernetes?

rokroskar:
@leafty is there other stuff to do here or can the `[WIP]` be removed?

leafty:
The pods endpoints leaks info about all the user pods, so I don't know if you want to modify it before merging.

rokroskar:
I definitely don't want to keep it there forever. It might be useful for debugging while we implement the other parts that need to watch the pods.

leafty:
Thx will swash and merge

PR #21, opened 2018-07-06
image pull secrets

leafty:
This PR adds the possibility to use a private registry for the `renku-notebooks` image.

Private image example:
https://gitlab.com/renku/renku-notebooks/container_registry
then:
```bash
$ helm upgrade -i renku-notebooks \
    --namespace test \
    --set image.repository="registry.gitlab.com/renku/renku-notebooks" \
    --set image.tag="pull-secrets" renku-notebooks \
    --set global.renku.domain=$(minikube ip)
```

You will see the image pull fail.
```bash
$ kubectl -n test create secret docker-registry regcred \
    --docker-server=registry.gitlab.com \
    --docker-username="gitlab+deploy-token-5965" \
    --docker-password="***" \
    --docker-email="***"
$ helm upgrade -i renku-notebooks \
    --namespace test \
    --set image.repository="registry.gitlab.com/renku/renku-notebooks" \
    --set image.tag="pull-secrets" renku-notebooks \
    --set global.renku.domain=$(minikube ip) \
    --set "image.pullSecrets[0]=regcred"
```

---
addresses https://github.com/SwissDataScienceCenter/renku/issues/176 and https://github.com/SwissDataScienceCenter/renku/issues/276

### DIFF
--- a/helm-chart/renku-notebooks/templates/deployment.yaml
+++ b/helm-chart/renku-notebooks/templates/deployment.yaml
@@ -66,3 +66,9 @@ spec:
       tolerations:
 {{ toYaml . | indent 8 }}
     {{- end }}
+      {{- if .Values.image.pullSecrets }}
+      imagePullSecrets:
+      {{- range .Values.image.pullSecrets }}
+        - name: {{ . }}
+      {{- end}}
+      {{- end }}
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -9,6 +9,13 @@ image:
   tag: latest
   pullPolicy: IfNotPresent
 
+  ## Optionally specify an array of imagePullSecrets.
+  ## Secrets must be manually created in the namespace.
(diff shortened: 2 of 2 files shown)


rokroskar:
Great, thanks for picking this one up -- but don't we need to create this secret per user and potentially per request?

leafty:
This is to be able to make gitlab.com/renku/stage-renku private.

This doesn't affect user JH pods, but yes either you deploy a sudo secret or you manage one for every user.

rokroskar:
So the lazy solution is to use a sudo secret... but eventually we want to handle the image pull secrets per user I suppose. 

Can adding the sudo secret be automated? Maybe via another post-install script? Ideally this wouldn't be a manual step.

leafty:
This can be done in a post install step, but I will also have to check if the JH chart gives the ability to read secrets to the hub.

rokroskar:
@leafty you mean to pass secrets to the pods? That's done via the `image_pull_secrets` in the spawner: https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L400

PR #22, opened 2018-07-09
Added capability to use private images from gitlab

leafty:
Requires: https://github.com/SwissDataScienceCenter/renku/pull/301

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -178,7 +178,10 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
         'project': project,
     }
     if os.environ.get('GITLAB_REGISTRY_SECRET'):
-        payload['image_pull_secrets'] = os.environ['GITLAB_REGISTRY_SECRET']
+        payload['image_pull_secrets'] = payload.get('image_pull_secrets', [])
+        payload['image_pull_secrets'] = payload['image_pull_secrets'].append(
+            os.environ['GITLAB_REGISTRY_SECRET']
+        )
     r = requests.request(
         'POST',
         '{prefix}/users/{user[name]}/servers/{server_name}'.format(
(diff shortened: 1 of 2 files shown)


src/notebooks_service.py
```diff
@@ -170,18 +170,32 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
...
     # 1. launch using spawner that checks the access
+    payload = {
+        'branch': request.args.get('branch', 'master'),
+        'commit_sha': commit_sha,
+        'namespace': namespace,
+        'notebook': notebook,
+        'project': project,
+    }
+    if os.environ.get('GITLAB_REGISTRY_SECRET'):
+        payload['image_pull_secrets'] = payload.get('image_pull_secrets', [])
```
rokroskar: `payload` here will never contain `image_pull_secrets` -- or is that something you expect in the future?

src/notebooks_service.py
```diff
@@ -170,18 +170,32 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
...
+        payload['image_pull_secrets'].append(
+            os.environ['GITLAB_REGISTRY_SECRET']
+        )
+    app.logger.debug(
+        'POST {prefix}/users/{user[name]}/servers/{server_name}'.format(
+            prefix=auth.api_url, user=user, server_name=server_name
+        )
+    )
+    app.logger.debug(
+        'data = {}'.format(payload)
```
rokroskar: probably don't want to log the image registry secret

src/notebooks_service.py
leafty: I thought it would make the code robust to changes anywhere.

rokroskar: 👍 just making sure I'm not missing something obvious

src/notebooks_service.py
leafty: It's just the names of the secret like 'renku-notebooks-registry' so not really critical.

rokroskar: ah true

Issue #24, opened 2018-07-10
Create a better server launch flow


rokroskar:
At the moment, we simply issue an API request to jupyterhub for a server spawn and wait until jupyterhub reports that this server is running. This is definitely not the way we want to handle server launches. 

We should improve the sequence to provide some extra feedback to the user about what is happening. Most importantly, the notebooks service flask app should not block waiting for the server to spawn. We should discuss how to move this forward -- some options: 

* if the server for `<namespace>/<project>/<sha>` is not running, the endpoint should return an page that shows the status of the pod/server being launched and a 202 return code 
* if the server for `<namespace>/<project>/<sha>` is running, the server/notebook is returned as normal
* the page with the status should redirect to the running server once it's up
* we should use k8s/docker clients for checking on the server spawn so we can report errors as they arise. If the pod launch errors, call `DELETE` on the JH API for the server so that it is immediately removed from the proxy etc. 
* ~~do we need a `<namespace>/<project>/<sha>/status` endpoint? Or just `<namespace>/<project>/<sha>?status`?~~ No, use CRUD and make `POST` launch the notebook, `GET` gives the status of the launch

Once this is done, it will most likely fix #23

leafty:
@rokroskar Closed via #25?

rokroskar:
yep.

PR #25, opened 2018-07-11
make the main endpoint more CRUD

rokroskar:
- [x] use POST for launch and GET for server status
- [x] give some more user-friendly feedback (simple auto-reload in html?)
- [x] basic use of k8s API to get notebook container status

addresses #24

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,11 +8,9 @@ docker = "*"
 jupyterhub = "*"
 python-gitlab = "*"
 requests-oauthlib = "*"
-Flask = "*"
+flask = "*"
 escapism = "*"
 kubernetes = "*"
 
 [dev-packages]
-
-[requires]
-python_version = "3.6"
+chartpress = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -14,3 +14,6 @@ kubernetes = "*"
 
 [dev-packages]
 chartpress = "*"
+
+[requires]
+python_version = "3.6"
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -71,6 +71,7 @@ class ReverseProxied(object):
 
     :param app: the WSGI application
     """
+
     def __init__(self, app):
         self.app = app
 
@@ -134,7 +135,9 @@ def authenticated(f):
         else:
             # redirect to login url on failed auth
             state = auth.generate_state(next_url=request.path)
-            app.logger.debug('Auth flow, redirecting to: {}'.format(auth.login_url))
+            app.logger.debug(
+                'Auth flow, redirecting to: {}'.format(auth.login_url)
+            )
             response = make_response(
                 redirect(auth.login_url + '&state=%s' % state)
             )
@@ -192,15 +195,15 @@ def user_server_is_running(user, server_name):
     headers = {auth.auth_header_name: 'token {0}'.format(auth.api_token)}
     user_info = requests.request(
         'GET',
-        '{prefix}/users/{user[name]}'.format(
-            prefix=auth.api_url, user=user
-        ),
+        '{prefix}/users/{user[name]}'.format(prefix=auth.api_url, user=user),
         headers=headers
     ).json()
     return server_name in user_info['servers']
 
 
-def get_user_server_status(user, namespace, project, commit_sha, notebook=None):
+def get_user_server_status(
+    user, namespace, project, commit_sha, notebook=None
+):
     """Returns the current status of a user named server or redirect to it if running"""
     server_name = _server_name(namespace, project, commit_sha)
     notebook_url = _notebook_url(user, server_name, notebook)
@@ -233,7 +236,9 @@ def get_user_server_status(user, namespace, project, commit_sha, notebook=None):
 @authenticated
 def notebook_status(user, namespace, project, commit_sha, notebook=None):
     """Returns the current status of a user named server or redirect to it if running"""
-    return get_user_server_status(user, namespace, project, commit_sha, notebook)
+    return get_user_server_status(
+        user, namespace, project, commit_sha, notebook
+    )
 
 
 @app.route(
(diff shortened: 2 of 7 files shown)


rokroskar:
The basic functionality is there -- needs to be extended to handle e.g. failed pod launches

src/notebooks_service.py
```diff
@@ -29,14 +29,24 @@
...
 
 SERVICE_PREFIX = os.environ.get('JUPYTERHUB_SERVICE_PREFIX', '/')
 """Service prefix is set by JupyterHub service spawner."""
 
+ANNOTATION_PREFIX = 'hub.jupyter.org'
+"""The prefix used for annotations by the KubeSpawner."""
+
+# check if we are running on k8s
+try:
+    from kubernetes import client, config
```
leafty: Not sure this test is enough. Running:
```python
config.load_incluster_config()
```
should fail outside of k8s.
You can also see if paths like `/var/run/secrets/kubernetes.io/serviceaccount` exists and are not empty.

src/notebooks_service.py
```diff
@@ -132,32 +155,80 @@ def whoami(user):
...
+            return redirect(notebook_url)
+
+        elif r.status_code == 202:
+            # server is spawning - return a 202 with a location header
+            app.logger.debug(
+                'spawn initialized for {server_name}'.format(
+                    server_name=server_name
+                )
+            )
+            status = 'spawn initialized'
```
leafty: is it intended that we don't return here?

src/notebooks_service.py
```diff
@@ -169,63 +240,16 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
...
-        ),
-        json=payload,
-        headers=headers,
-    )
-
-    # 2. redirect to the server when ready
-    if r.status_code == 201:
-        # server is already running
-        return redirect(notebook_url)
+    status = get_notebook_container_status(user['name'], server_name)
```
leafty: Is that overwritten?
```python
status = 'spawn initialized'
```

Pipfile
```diff
@@ -8,11 +8,9 @@ docker = "*"
...
 requests-oauthlib = "*"
-Flask = "*"
+flask = "*"
 escapism = "*"
 kubernetes = "*"
 
 [dev-packages]
-
-[requires]
-python_version = "3.6"
```
leafty: Removed because it's forced by the Dockerfile?

src/notebooks_service.py
rokroskar: oops yes this status is superflous. If we return here, it wouldn't make sense to return the html, so I just log that the spawn has been initiated and when we get to the bottom of the function we return with the html that waits for the server.

Pipfile
rokroskar: not sure actually...

src/notebooks_service.py
```diff
@@ -132,32 +160,79 @@ def whoami(user):
...
+    notebook_url = _notebook_url(user, server_name, notebook)
     headers = {auth.auth_header_name: 'token {0}'.format(auth.api_token)}
 
-    notebook_url = urljoin(
-        os.environ.get('JUPYTERHUB_BASE_URL'),
-        'user/{user[name]}/{server_name}/'.format(
-            user=user, server_name=server_name
+    status = 'not started'
+
+    if request.method == 'POST':
```
leafty: This if block is too long (readability of code)

leafty:
@jirikuncar Since I authored a sizeable bit too, can I get you to review this? Would help to have another pair of eyes look over this.

leafty:
@rokroskar can you check my changes? it's not a small change

src/notebooks_service.py
```diff
@@ -29,14 +29,29 @@
...
+# check if we are running on k8s
+try:
+    from kubernetes import client, config
+    config.load_incluster_config()
+    with open(
+        '/var/run/secrets/kubernetes.io/serviceaccount/namespace', 'rt'
+    ) as f:
+        kubernetes_namespace = f.read()
+    KUBERNETES = True
+except:
```
jirikuncar: Please do not use bare except.

src/templates/server_status.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
```
jirikuncar: Templates should contain a license header too.

src/notebooks_service.py
```diff
@@ -29,14 +29,29 @@
...
+    KUBERNETES = False
```
jirikuncar: Does it make sense to start the server without K8S?

src/templates/server_status.html
leafty: This is currently not the case in renku-ui --> https://github.com/SwissDataScienceCenter/renku-ui/blob/master/public/index.html

Also, I do not see putting a license header on HTML pages as a standard practice. (Check sites like github, docker, gitlab, stack overflow, etc.)

jirikuncar: Since it's a Jinja2 template you can include it as a comment.

```
{#
...
#}
```

src/notebooks_service.py
```diff
@@ -199,33 +282,25 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
...
-    # 2. redirect to the server when ready
+    # 2. check response, we expect:
+    #   - HTTP 201 if the server is already running; in this case redirect to it
+    #   - HTTP 202 if the server is spawning
     if r.status_code == 201:
-        # server is already running
+        app.logger.debug(
+            'server {server_name} running'.format(server_name=server_name)
+        )
         return redirect(notebook_url)
```
rokroskar: we don't really need to return here since `get_user_server_status` will redirect anyway?

src/templates/server_status.html
rokroskar: is it necessary? required? @erbou ?

rokroskar:
I think this is gtg @jirikuncar and @leafty -- just waiting on @ableuler to do two minor changes to the UI PR https://github.com/SwissDataScienceCenter/renku-ui/pull/266

leafty:
Closes #26

PR #32, opened 2018-08-16
add telepresence script

rokroskar:
make it possible to use telepresence to develop the notebooks service component


.gitignore
```diff
@@ -252,3 +252,5 @@ tags
 
 
 # End of https://www.gitignore.io/api/vim,linux,emacs,macos,python,sublimetext
+
+telepresence.log
```
leafty: This is conspicuous, add it before the gitignore.io section :wink:

rokroskar: why? You mean on top of the file?

PR #33, opened 2018-08-16
Status handling

rokroskar:
Handle the situation where a user tries to re-start a server that is in the process of shutting down. 

todo:

- [x] template app secret in the chart

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -29,7 +29,7 @@ import docker
 import escapism
 import gitlab
 import requests
-from flask import Flask, Response, abort, make_response, redirect, render_template, request
+from flask import Flask, Response, abort, make_response, redirect, render_template, request, session
 from flask import jsonify
 from flask import send_from_directory
 from jupyterhub.services.auth import HubOAuth
@@ -91,6 +91,8 @@ class ReverseProxied(object):
 
 app = Flask(__name__)
 app.wsgi_app = ReverseProxied(app.wsgi_app)
+
+app.secret_key = os.environ.get('NOTEBOOKS_SECRET_KEY', '12345abcde')
 
 
 def _server_name(namespace, project, commit_sha):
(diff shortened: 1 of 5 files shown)


rokroskar:
PR is staged, seems to work as expected but please test.

src/templates/server_status.html
```diff
@@ -28,11 +28,26 @@
...
+  <meta http-equiv="refresh" content="2">
+
+  {% elif status == "stopping" %}
+  <h3>The server is in the process of terminating.</h3>
+  <meta http-equiv="refresh" content="2">
+  {% endif %}
 </div>
 
-<meta http-equiv="refresh" content="2">
+<script>
```
leafty: Invalid html here

rokroskar: oops! thanks for noticing that!

src/notebooks_service.py
```diff
@@ -92,6 +92,8 @@ def __call__(self, environ, start_response):
 app = Flask(__name__)
 app.wsgi_app = ReverseProxied(app.wsgi_app)
 
+app.secret_key = os.environ.get('NOTEBOOKS_SECRET_KEY', '12345abcde')
```
ableuler: Maybe provide default through charts, not at the application level.

src/templates/server_status.html
```diff
@@ -28,11 +28,24 @@
 
 <div class="jumbotron text-center">
   <h1>Server for {{namespace}}/{{project}}/{{commit_sha}} status: {{status}}</h1>
-  <p>This page will automatically refresh and redirect you to the server when it is running.</p>
-</div>
+  {% if status == "not found" %}
```
ableuler: For the moment this is a good improvement, but moving forward, the trick with the previous state in a cookie should be replaced with some server-side state handling (use the existing redis pod?). From a client side, I am just interested in the state of the server, and I should get a response for it without assuming that it was the same client which has triggered a certain action beforehand.

PR #34, opened 2018-08-17
use gunicorn as the http server

### LINKED ISSUE #31: do not use the flask development server
Use gunicorn or similar in the Docker image.

rokroskar:
closes #31

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,10 +19,14 @@ FROM python:3.6-alpine
 MAINTAINER renku
 
 RUN apk update && \
-    apk add --no-cache curl && \
+    apk add --no-cache curl build-base libffi-dev openssl-dev && \
+    pip install -U pip && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -19,11 +19,15 @@ FROM python:3.6-alpine
 MAINTAINER renku
 
 RUN apk update && \
-    apk add --no-cache curl && \
+    apk add --no-cache curl build-base libffi-dev openssl-dev && \
```
ableuler: I think some of these dependencies are actually not necessary and ended up weeding out the Dockerfile where you probably got them from (see https://github.com/SwissDataScienceCenter/renku-gateway/pull/43).

rokroskar: They are for installing dependencies for one of the cryptography libraries I think. There is no wheel for alpine so they have to compile from source.

PR #39, opened 2018-08-23
improve RESTfulness

rokroskar:
`POST` returns json as appropriate
`GET` returns json if `application/json` is requested, otherwise returns html. This is to facilitate tighter integration with the UI.


rokroskar:
put this up on staging, seems to behave as expected

rokroskar:
@ableuler good point, done in 7759038

PR #40, opened 2018-08-30
move all jupyter-related parts to this repository

### LINKED ISSUE #3: Use JH as a dependency
Jupyterhub should be a dependency of renku-notebooks.

rokroskar:
Ripping out all jupyter/jupyterhub related pieces from the renku repo to here. 

Note that it is now possible to deploy the notebooks service as a standalone service with a jupyterhub against e.g. gitlab.com. 

closes #3


helm-chart/renku-notebooks/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Notebooks service
 name: renku-notebooks
-version: 0.1.0
+version: 0.0-unclean-fd4775a
```
leafty: Chartpress caught you

helm-chart/renku-notebooks/templates/ingress.yaml
```diff
@@ -35,5 +35,9 @@ spec:
             backend:
               serviceName: {{ $fullName }}
               servicePort: http
+          # - path: /jupyterhub
+          #   backend:
+          #     serviceName: proxy-public
+          #     servicePort: 8000
```
leafty: You should probably uncomment this

helm-chart/renku-notebooks/values.yaml
```diff
@@ -2,11 +2,14 @@
...
 
+global:
+  useHTTPS: false
+
 replicaCount: 1
 
 image:
   repository: renku/renku-notebooks
-  tag: 'latest'
+  tag: '0.0-unclean-fd4775a'
```
leafty: Chartpress caught you

helm-chart/renku-notebooks/values.yaml
```diff
@@ -37,16 +41,6 @@ rbac:
...
-  annotations: {}
-    # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
-  path: /
-  hosts:
-    - chart-example.local
-  tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
```
leafty: Don't remove this

helm-chart/renku-notebooks/values.yaml
```diff
@@ -65,3 +59,97 @@ nodeSelector: {}
...
 affinity: {}
+
+## Configuration for the jupyterhub service
+jupyterhub:
+  rbac:
+    enabled: true
+  hub:
+    image:
+      name: renku/jupyterhub-k8s
+      tag: '0.0-unclean-fd4775a'
```
leafty: Chartpress caught you

helm-chart/renku-notebooks/values.yaml
```diff
@@ -65,3 +59,97 @@ nodeSelector: {}
...
+  prePuller:
+    hook:
+      enabled: false
+  auth:
+    type: gitlab
+    gitlab:
+      # Fill in the clientId and clientSecret after creating the OAuth app
+      # clientId: 1234
+      # clientSecret: 1234
+      callbackUrl: '""' # Forces the default callback url
```
leafty: clientid is missing, compared to values from https://github.com/SwissDataScienceCenter/renku

helm-chart/renku-notebooks/values.yaml
```diff
@@ -65,3 +59,97 @@ nodeSelector: {}
...
+  proxy:
+    service:
+      type: ClusterIP
+      nodePorts: {}
+    https:
+      enabled: false
+  cull:
+    enabled: true
+    timeout: 600
```
leafty: Keep old long timeout

helm-chart/renku-notebooks/values.yaml
```diff
@@ -65,3 +59,97 @@ nodeSelector: {}
...
+
+      # JupyterHub uses GITLAB_HOST which is a misnomer -- we use GITLAB_URL
+      # everywhere so set this mapping here
+      os.environ['GITLAB_HOST'] = os.getenv('GITLAB_URL')
+
+      #: Automatically begin the login process without showing the button.
+      c.Authenticator.auto_login = True
+
+      #: Enable named-servers
+      c.JupyterHub.allow_named_servers = True
```
leafty: Apparently, it's a chart value now, so I would remove this and use the value

helm-chart/renku-notebooks/templates/deployment.yaml
```diff
@@ -35,13 +35,17 @@ spec:
...
+              value: {{ .Values.jupyterhub_api_url }}
+            {{ else }}
               value: {{ template "notebooks.http" . }}://{{ .Values.global.renku.domain }}{{ .Values.jupyterhub_base_url }}hub/api
+            {{ end }}
             - name: JUPYTERHUB_SERVICE_PREFIX
               value: {{ .Values.jupyterhub_base_url }}services/notebooks/
             - name: JUPYTERHUB_BASE_URL
               value: {{ .Values.jupyterhub_base_url }}
-            - name: GITLAB_REGISTRY_SECRET
-              value: {{ template "renku.fullname" . }}-registry
```
leafty: Why remove this?

leafty:
It didn't work on staging, probably because some values in `jupyterhub.auth` were moved around:
```
traitlets.traitlets.TraitError: The 'client_id' trait of a GitLabOAuthenticator instance must be a unicode string, but a value of None <class 'NoneType'> was specified.
```

src/notebooks_service.py
```diff
@@ -330,8 +326,8 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
 
 @app.route(
     urljoin(SERVICE_PREFIX, '<namespace>/<project>/<commit_sha>'),
-    methods=['DELETE']
-)
+    methods=['DELETE'
+])
```
leafty: Is that intentional? It looks weird

helm-chart/renku-notebooks/templates/deployment.yaml
rokroskar: you're right, it shouldn't be removed -- not sure how to access secrets from the top-level chart in a sub-chart...

leafty: I don't think it needs to be modified

rokroskar: but if it is deployed standalone, the secret doesn't exist.

helm-chart/renku-notebooks/templates/ingress.yaml
rokroskar: well, the ingress is handled by renku. And when you run from minikube, it doesn't use the ingress, just a `NodePort`.

helm-chart/renku-notebooks/templates/deployment.yaml
leafty: ah ok. It is created manually at the moment (TODO: doc :stuck_out_tongue_closed_eyes: ) so nothing new here

rokroskar: yeah at the moment we have no way of accessing private repos on a gitlab instance that we do not control. This will require a) handling user secrets in the notebook app or b) somehow getting tokens with repo access from e.g. the gateway

PR #50, opened 2018-09-17
spawner: avoid infinite loop

### LINKED ISSUE #49: Introduce timeout for pending images.
Avoid infinite loop if build job is stuck.

See:
https://github.com/SwissDataScienceCenter/renku-notebooks/blob/d7ac292a4e9514a63ae7723f609af432e6c259e1/jupyterhub/spawners.py#L137-L168
was https://github.com/SwissDataScienceCenter/renku/issues/318 reported by @leafty

rokroskar:
closes #49

### DIFF
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -89,6 +89,9 @@ jupyterhub:
     extraEnv:
       DEBUG: 1
       JUPYTERHUB_SPAWNER_CLASS: spawners.RenkuKubeSpawner
+      ## timeout for waiting on an image build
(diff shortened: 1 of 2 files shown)


helm-chart/renku-notebooks/values.yaml
```diff
@@ -9,7 +9,7 @@ replicaCount: 1
 
 image:
   repository: renku/renku-notebooks
-  tag: 'latest'
+  tag: '0.latest'
```
leafty: What changes here?

rokroskar: oops, thanks for spotting that -- fixed

PR #51, opened 2018-09-18
jh: do not force lowercase

### LINKED ISSUE #45: Project with upper case is not started properly
The mounted volume (empty for) does not respect case at the moment. It results in a lab instance which cannot be used

Edit: This happens when converting existing projects to Renku (and not when creating them through the UI).

rokroskar:
closes #45

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -466,8 +466,7 @@ try:
                 username=safe_username[:10],
                 legacy_escape_username=legacy_escaped_username[:10],
                 servername=servername
-            ).lower()
-
+            )
             # just to be sure, still trim to 63 characters
             return rendered[:63]
 


rokroskar:
@leafty could be that the `self.name` has an uppercase in it -- let me check

rokroskar:
I tried with a capitalized namespace and capitalized project -- seems ok.

rokroskar:
I know what you mean - all the other variables are escaped in that piece of code except for self.name - but it looks like that must be escaped somewhere else.

leafty:
Grrr! I asked you if it was safe:
```
Pod \"jupyter-johann-2Et-johann-2et-newnew-a342109\" is invalid: [metadata.name: Invalid value: \"jupyter-johann-2Et-johann-2et-newnew-a342109\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```

rokroskar:
that's not good

PR #54, opened 2018-09-19
Fixing #45 again

rokroskar:
This commit got left off the previous PR.


rokroskar:
Tested on staging with `test.user` and `Test.Project`-- seems to yield correct pod names and preserves capitalization in the cloned project.

rokroskar:
careful it still might be broken, in which case the pouting may need to be longer.

PR #56, opened 2018-09-21
Refactor image build check 

rokroskar:
Move the logic regarding image selection out of the spawner and into the notebook service. The spawner should just launch whichever image is given to it.

### DIFF
--- a/helm-chart/renku-notebooks/templates/deployment.yaml
+++ b/helm-chart/renku-notebooks/templates/deployment.yaml
@@ -28,34 +28,36 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           env:
+            - name: DEFAULT_NOTEBOOK_IMAGE
+              value: "{{ .Values.jupyterhub.singleuser.image.name }}:{{ .Values.jupyterhub.singleuser.image.tag }}"
             {{ if eq .Values.debug true }}
             - name: FLASK_DEBUG
               value: "1"
             {{ end }}
             - name: JUPYTERHUB_API_TOKEN
-              value: {{ .Values.jupyterhub_api_token }}
+              value: {{ .Values.jupyterhub.hub.services.notebooks.api_token }}
             - name: JUPYTERHUB_API_URL
-            {{ if .Values.jupyterhub_api_url }}
-              value: {{ .Values.jupyterhub_api_url }}
-            {{ else }}
-              value: {{ template "notebooks.http" . }}://{{ .Values.global.renku.domain }}{{ .Values.jupyterhub_base_url }}hub/api
-            {{ end }}
+              value: {{ template "notebooks.http" . }}://{{ .Values.global.renku.domain }}{{ .Values.jupyterhub.hub.baseUrl }}hub/api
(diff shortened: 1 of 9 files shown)


rokroskar:
This seems to work, tested on staging. Would be good to have a second pair of eyes.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -89,8 +96,6 @@ jupyterhub:
     extraEnv:
       DEBUG: 1
       JUPYTERHUB_SPAWNER_CLASS: spawners.RenkuKubeSpawner
-      ## timeout for waiting on an image build
-      # GITLAB_IMAGE_BUILD_TIMEOUT: 600
```
leafty: Need to move it instead, no?

leafty:
I see. I guess it might be good to rethink the PR then. Let's see how we can have public image selection together with repo-based images.

rokroskar:
@leafty I am happy to remove that part for now as it doesn't really address the point of this PR in any case. We may want to look at the JupyterHub spawner `options_form` feature which allows you to do something along these lines and control which images are allowed.

helm-chart/renku-notebooks/templates/deployment.yaml
```diff
@@ -48,6 +48,14 @@ spec:
               value: {{ .Values.jupyterhub.hub.services.notebooks.oauth_client_id }}
             - name: GITLAB_REGISTRY_SECRET
               value: {{ template "renku.fullname" . }}-registry
+            - name: GITLAB_URL
+            {{ if .Values.gitlab_url }}
```
jirikuncar: `.Values.gitlab.url` (see https://github.com/SwissDataScienceCenter/renku-notebooks/pull/56/files#diff-4a2cd40c069f23f7cd71ab6a74028709R10)?

rokroskar:
@leafty can you please review again?

helm-chart/renku-notebooks/templates/deployment.yaml
```diff
@@ -28,26 +28,36 @@ spec:
...
           env:
+            - name: DEFAULT_NOTEBOOK_IMAGE
+              value: "{{ .Values.jupyterhub.singleuser.image.name }}:{{ .Values.jupyterhub.singleuser.image.tag }}"
             {{ if eq .Values.debug true }}
             - name: FLASK_DEBUG
               value: "1"
             {{ end }}
             - name: JUPYTERHUB_API_TOKEN
-              value: {{ .Values.jupyterhub_api_token }}
+              value: {{ .Values.jupyterhub.hub.services.notebooks.api_token }}
```
leafty: :+1:

singleuser/Dockerfile
```diff
@@ -1,12 +1,12 @@
 # Build as renku/singleuser
 # Run with the DockerSpawner in JupyterHub
 
-ARG JUPYTERHUB_VERSION=0.9
+ARG JUPYTERHUB_VERSION=0.9.2
```
leafty: they made `0.9.1` disappear, will this one last?

rokroskar: it's there: https://github.com/jupyterhub/jupyterhub/releases/tag/0.9.1

PR #64, opened 2018-10-01
remove the custom git extension installed in renku/singleuser

leafty:
Supersedes https://github.com/SwissDataScienceCenter/renku/pull/373


leafty:
Can be tried with -> https://renkulab.io/gitlab/johann-michael.thiebaut/test-git-extension

PR #73, opened 2018-10-18
authenticate with jh token in request header

### LINKED ISSUE #72: allow authorization through token in the header
The `@authenticated` decorator should accept JupyterHub-issued tokens.

rokroskar:
closes #72

### DIFF
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -134,7 +134,7 @@ def authenticated(f):
         token = request.cookies.get(
             auth.cookie_name
         ) or request.headers.get('Authorization',
-                                 'token').split('token')[1].strip()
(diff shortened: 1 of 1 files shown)


src/notebooks_service.py
```diff
@@ -131,7 +131,10 @@ def authenticated(f):
 
     @wraps(f)
     def decorated(*args, **kwargs):
-        token = request.cookies.get(auth.cookie_name)
+        token = request.cookies.get(
+            auth.cookie_name
+        ) or request.headers.get('Authorization',
+                                 'token').split('token')[1].strip()
```
leafty:
```suggestion
                                 'token').split('token', 1)[1].strip()
```

PR #76, opened 2018-10-18
skip smudge by default and provide server_options override

### LINKED ISSUE #74: Do not load LFS objects by default
We should have `GIT_LFS_SKIP_SMUDGE=1` set by default and the possibility to override it via `server_options`.

rokroskar:
closes #74

### DIFF
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -82,29 +82,30 @@ tolerations: []
 affinity: {}
 
 serverOptions:
-  resources:
-    cpu_request:
-      displayName: Number of CPUs
-      type: integer
-      default: 1
-      range: [1, 8]
-    mem_request:
-      displayName: Memory
-      type: enum
-      default: 2G
-      options: [2G, 4G, 8G]
-    gpu:
-      displayName: Number of GPUs
-      type: integer
-      default: 0
-      range: [0, 4]
+  cpu_request:
+    displayName: CPU request
+    type: enum
+    default: 0.1
+    options: [0.1, 0.5, 1.0, 2.0, 4.0]
+  mem_request:
+    displayName: Memory request
+    type: enum
+    default: 1G
+    options: [1G, 2G, 4G]
+  gpu_request:
+    displayName: GPU request
+    type: int
+    default: 0
+    range: [0, 4]
   defaultUrl:
+    displayName: default landing URL of the notebook server
+    type: string
     default: /lab
     options: [/lab]
-  lfs_skip_smudge:
-    default: 1
-    options: [0,1]
-    displayName: Do not automatically fetch data
+  lfs_auto_fetch:
+    displayName: Automatically fetch LFS data
+    type: boolean
+    default: false
 
 ## Configuration for the jupyterhub service
 jupyterhub:
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -101,6 +101,10 @@ serverOptions:
   defaultUrl:
     default: /lab
     options: [/lab]
+  lfs_skip_smudge:
+    default: 1
+    options: [0,1]
+    displayName: Do not automatically fetch data
 
 ## Configuration for the jupyterhub service
 jupyterhub:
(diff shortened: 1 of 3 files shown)


rokroskar:
Yep you're absolutely right - we could just make the default that nothing gets pulled from LFS.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -101,6 +101,10 @@ serverOptions:
   defaultUrl:
     default: /lab
     options: [/lab]
+  lfs_skip_smudge:
```
ableuler: These changes won't be picked up by the UI. So far I assumed that the top-level properties are only `defaultUrl` and `resources`, and I display all options under `resources` automatically. If you really want this to be a configurable option through the UI, add the `enum` type to the `lsf_skip_smudge` object and move it into `resources`. In general I would prefer a flat structure for the `server_options` object as long as the total number of options is manageable.

rokroskar: what about 

```
serverOptions:
  cpu_request:
    displayName: Number of CPUs
    type: integer
    default: 1
    range: [1, 8]
  mem_request:
    displayName: Memory
    type: enum
    default: 2G
    options: [2G, 4G, 8G]
  gpu:
    displayName: Number of GPUs
    type: integer
    default: 0
    range: [0, 4]
  defaultUrl:
    default: /lab
    options: [/lab]
  lfs_skip_smudge:
    default: 1
    options: [0,1]
    displayName: Do not automatically fetch data
```

ableuler: Yes, this would make the ui a bit simpler.

rokroskar: I had `resources` in there initially to be a bit closer to the k8s pod spec, but I need to do the mapping by hand anyway so it's not very useful.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -82,25 +82,30 @@ tolerations: []
...
+  cpu_request:
+    displayName: CPU request
+    type: enum
+    default: 0.1
+    options: [0.1, 0.5, 1.0, 2.0, 4.0]
+  mem_request:
+    displayName: Memory request
+    type: enum
+    default: 1G
+    options: [1G, 2G, 4G]
```
leafty: `1G` or `1Gi`? I like my power of teo :wink:

helm-chart/renku-notebooks/values.yaml
```diff
@@ -82,25 +82,30 @@ tolerations: []
...
+  gpu_request:
+    displayName: GPU request
+    type: int
+    default: 0
+    range: [0, 4]
```
leafty: Maybe have `[0 , 1]` as the default range?

src/notebooks_service.py
```diff
@@ -133,8 +133,8 @@ def authenticated(f):
     def decorated(*args, **kwargs):
         token = request.cookies.get(
             auth.cookie_name
-        ) or request.headers.get('Authorization',
-                                 'token').split('token', 1)[1].strip()
+        ) or request.headers.get('Authorization', 'token').split('token',
+                                                                 1)[1].strip()
```
leafty: Will fail on headers like `Authorization: ehskdfbklshfuiw` with http 500. We can merge and create an issue just after.

helm-chart/renku-notebooks/values.yaml
leafty:
```suggestion
    range: [0, 1]
```

helm-chart/renku-notebooks/values.yaml
```diff
@@ -99,7 +99,7 @@ serverOptions:
     range: [0, 1]
   defaultUrl:
     displayName: default landing URL of the notebook server
-    type: string
+    type: enum
```
leafty: why? what if someone has a new thing like r shiny?

ableuler: We can define a "string" option (not implemented yet on the ui side), but then the options don't make sense. The original plan was to provide options that would translate into a dropdown.

rokroskar: `enum` is correct, this was my mistake

PR #82, opened 2018-10-19
Use symlink for renku instead of alias

### LINKED ISSUE #81: Use symlink, not alias for renku in singleuser
https://github.com/SwissDataScienceCenter/renku-notebooks/blob/e3a5b98fe30d9a1cdcf5daf4bf1c6d1cf5dadd5e/singleuser/Dockerfile#L51

From magic recipe:
```bash
conda create -y -n renku python=3.6
$(conda env list | grep renku | awk '{print $2}')/bin/pip install -e git+https://github.com/SwissDataScienceCenter/renku-python.git#egg=renku
mkdir -p ~/.renku/bin
ln -s "$(conda env list | grep renku | awk '{print $2}')/bin/renku" ~/.renku/bin/renku
echo "export PATH=~/.renku/bin:$PATH" >> $HOME/.bashrc
source $HOME/.bashrc
renku --version
which renku
```

leafty:
Closes #81

### DIFF
--- a/singleuser/Dockerfile
+++ b/singleuser/Dockerfile
@@ -48,7 +48,9 @@ RUN python3 -m pip install -U pip && \
 # install renku-python
 RUN conda create -y -n renku python=3.6 && \
     $CONDA_DIR/envs/renku/bin/pip install renku==$RENKU_VERSION && \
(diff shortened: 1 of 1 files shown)


leafty:
@rok please reapprove. it did not need the awk stuff here

PR #87, opened 2018-10-26
Update chartpress

rokroskar:
remove the dependency of image builds on the entire `renku-notebooks` directory. This should result in fewer image builds.


rokroskar:
@leafty I have no idea, but let's try :) 

Actually I was worried about the fact that you rely on commits matching the chart version in your CD pipeline. But I think in this case, none of the tags affected here fall into that category.

leafty:
Yes, so doing that for singleuser images should be save.

PR #94, opened 2018-11-02
disable smudge globally

### LINKED ISSUE #93: disable smudge with repo option
This should be set in the repository on checkout: 

```
git lfs install --skip-smudge --local
```

Otherwise, every `git checkout` command will try to pull LFS objects leading to general unhappiness.

rokroskar:
Install LFS globally and disable or enable smudge based on options. 

addresses https://github.com/SwissDataScienceCenter/renku/issues/343
closes #93

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -325,24 +325,24 @@ try:
                 image='alpine/git:latest',
                 command=['sh', '-c'],
                 args=[
-                    'export GIT_LFS_SKIP_SMUDGE={lfs_skip_smudge} && '
                     'rm -rf {mount_path}/* && '
                     '(rm -rf {mount_path}/.* || true) && '
                     'apk update && apk add git-lfs && '
+                    'git lfs --install {lfs_skip_smudge} --global && '
                     'git clone {repository} {mount_path} && '
+                    'git lfs --install {lfs_skip_smudge} --local && '
                     '(git checkout {branch} || git checkout -b {branch}) && '
                     'git submodule init && git submodule update && '
                     'git reset --hard {commit_sha} &&'
-                    'git lfs install --local &&'
                     'git config push.default simple &&'
                     'chown 1000:100 -Rc {mount_path}'.format(
                         branch=options.get('branch', 'master'),
                         commit_sha=options.get('commit_sha'),
                         mount_path=volume_mount['mountPath'],
                         repository=repository,
-                        lfs_skip_smudge=int(
-                            not server_options.get('lfs_auto_fetch')
-                        )
+                        lfs_skip_smudge=''
+                        if server_options.get('lfs_auto_fetch') else
+                        '--skip-smudge',
                     )
                 ],
                 volume_mounts=[volume_mount],


rokroskar:
Ah you're right -- @jirikuncar I think the requirement of having git LFS installed explicitly in the present repo is a bit against how LFS works in general and may be a root of some of these issues. We should remove that IMHO.

jupyterhub/spawners.py
```diff
@@ -325,24 +325,24 @@ def get_pod_manifest(self):
                 image='alpine/git:latest',
                 command=['sh', '-c'],
                 args=[
-                    'export GIT_LFS_SKIP_SMUDGE={lfs_skip_smudge} && '
+                    'git lfs --install {lfs_skip_smudge} && '
```
leafty: Remove that line and then it's good to go

jupyterhub/spawners.py
```diff
@@ -325,24 +325,23 @@ def get_pod_manifest(self):
                 image='alpine/git:latest',
                 command=['sh', '-c'],
                 args=[
-                    'export GIT_LFS_SKIP_SMUDGE={lfs_skip_smudge} && '
                     'rm -rf {mount_path}/* && '
                     '(rm -rf {mount_path}/.* || true) && '
                     'apk update && apk add git-lfs && '
+                    'git lfs --install {lfs_skip_smudge} && '
```
leafty: `--local` :scream:

rokroskar: not needed anymore: https://github.com/SwissDataScienceCenter/renku-python/pull/347/files#diff-3d81f436074c1bb23a072c4663d07c18R363

leafty: Do we wipe all existing projects then? :wink:

rokroskar: so for backwards compatibility we need this line, plus a line with `--local` after the repo is cloned. 😭

rokroskar:
the git config is now:

```
file:/etc/gitconfig     filter.lfs.clean=git-lfs clean -- %f
file:/etc/gitconfig     filter.lfs.smudge=git-lfs smudge --skip -- %f
file:/etc/gitconfig     filter.lfs.process=git-lfs filter-process --skip
file:/etc/gitconfig     filter.lfs.required=true
file:.git/config        core.repositoryformatversion=0
file:.git/config        core.filemode=true
file:.git/config        core.bare=false
file:.git/config        core.logallrefupdates=true
file:.git/config        core.ignorecase=true
file:.git/config        remote.origin.url=https://github.com/SwissDataScienceCenter/renku.git
file:.git/config        remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
file:.git/config        branch.master.remote=origin
file:.git/config        branch.master.merge=refs/heads/master
file:.git/config        filter.lfs.required=true
file:.git/config        filter.lfs.clean=git-lfs clean -- %f
file:.git/config        filter.lfs.smudge=git-lfs smudge --skip -- %f
file:.git/config        filter.lfs.process=git-lfs filter-process --skip
```

PR #99, opened 2018-11-14
Refactor image builds 

rokroskar:
Moves image builds to https://github.com/SwissDataScienceCenter/renku-jupyter and fixes the documentation for local deployment.


rokroskar:
@leafty done - image builds seem to be working in `renku-jupyter` so I think this is gtg

PR #108, opened 2019-03-04
build(helm): extra config map with env for KubeSpawner

jirikuncar:
Adds option to include `SENTRY_DSN` configuration.


rokroskar:
Why does this need a configmap just to pass an environment variable? Can't you just pass the environment variable to the deployment directly? I guess you are trying to do this change entirely in the configuration and not in the code?

jirikuncar:
@rokroskar I wanted to avoid sharing the env variable with the hub itself.

PR #119, opened 2019-03-14
116 make git init image

rokroskar:


git-clone/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ "$LFS_AUTO_FETCH" = 1 ]; then LFS_SKIP_SMUDGE="";
+else LFS_SKIP_SMUDGE="--skip-smudge";
+fi
+
+rm -rf ${MOUNT_PATH}/*
+(rm -rf ${MOUNT_PATH}/.* || true)
+git lfs install $LFS_SKIP_SMUDGE --system
+git clone $REPOSITORY ${MOUNT_PATH}
+git lfs install $LFS_SKIP_SMUDGE --local
```
jirikuncar: can we use `GIT_LFS_SKIP_SMUDGE` which should work without using option?

rokroskar: true, that would be better

git-clone/Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG GIT_CLONE_VERSION=1.0.7
```
jirikuncar:
```suggestion
ARG GIT_ALPINE_VERSION =1.0.7
```

git-clone/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ARG BASE_IMAGE=alpine/git:$GIT_CLONE_VERSION
```
jirikuncar:
```suggestion
ARG BASE_IMAGE=alpine/git:$GIT_ALPINE_VERSION
```

git-clone/entrypoint.sh
rokroskar: Actually I changed my mind -- the reason for doing it this way is that it gets baked into the repository configuration. So I would leave it as-is.

rokroskar:
@jirikuncar can you please re-review?

git-clone/entrypoint.sh
jirikuncar: I don't get why having it baked into the repository is better.

rokroskar: because this is a volume that is shared between containers. We then have to pass the environment variable to all the relevant containers, which leaves more room for error. Here we just set it once. I know that passing an environment variable is not the end of the world, but errors with LFS configuration lead to potentially catastrophic repository... "entanglements" :)

PR #130, opened 2019-04-10
feat: added pre-stop post-start hooks

### LINKED ISSUE #96: Session backup
Sometimes notebook instances stop (e.g. node enters an out of memory condition) and current work is permanently lost.

Is there a way to provide some form of backup for these cases?

### LINKED ISSUE #126: Use container lifecycle hooks to save work
In order to avoid users losing work from their notebook sessions, we could save the repository into a backup branch when the container is being terminated. Lifecycle hooks can be added to the pod via the jupyterhub kubespawner: https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L710. See https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/ for background.

jachro:
This PR adds lifecycle hooks to the jupyterhub spawners.

closes #96 
closes #126

### DIFF
--- a/jupyterhub/requirements-k8s.txt
+++ b/jupyterhub/requirements-k8s.txt
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Installing from Git until a new version of kubespawner (introducing 'delete_grace_period') is released.
+-e git://github.com/jupyterhub/kubespawner.git@46a4b109c5e657a4c3d5bfa8ea4731ec6564ea13#egg=jupyterhub-kubespawner
 docker
 escapism
 python-gitlab
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -351,11 +351,6 @@ try:
             self.init_containers.append(init_container)
 
             self.lifecycle_hooks={
-                "postStart": {
-                    "exec": {
-                        "command": ["/bin/sh", "-c", "/usr/local/bin/post-start.sh"]
-                    }
-                },
                 "preStop": {
                     "exec": {
                         "command": ["/bin/sh", "-c", "/usr/local/bin/pre-stop.sh"]
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -353,7 +353,7 @@ try:
             self.lifecycle_hooks={
                 "preStop": {
                     "exec": {
-                        "command": ["/bin/sh", "-c", "/usr/local/bin/pre-stop.sh"]
+                        "command": ["/bin/sh", "-c", "/usr/local/bin/pre-stop.sh", "||", "true"]
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


jupyterhub/requirements-k8s.txt
```diff
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+-e git://github.com/jupyterhub/kubespawner.git@46a4b109c5e657a4c3d5bfa8ea4731ec6564ea13#egg=jupyterhub-kubespawner
```
rokroskar: can you add a comment here that installing from git is temporary until a new version of kubespawner is released?

jachro: Done

jupyterhub/spawners.py
```diff
@@ -350,6 +350,19 @@ def get_pod_manifest(self):
             )
             self.init_containers.append(init_container)
 
+            self.lifecycle_hooks={
+                "postStart": {
```
jirikuncar: I wonder if the functionality of the `postStart` could not be moved to the init container.

jachro: Unfortunately, it cannot. I tried that.

rokroskar: well you could run the script inside the init container, just not as a hook. However, one thing we discussed to extend the functionality here is to use the same script to notify the user in their shell about what has happened or if something went wrong to tell them where to find the saved session.

rokroskar: @jachro what's your take on this? conceptually is the autosave check different from the repo clone that happens in the init container somehow and should therefore be kept inside the notebook container? 

This actually speaks against it (from [the docs](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks)): 

"This hook executes immediately after a container is created. However, there is no guarantee that the hook will execute before the container ENTRYPOINT"

So if the script takes some time to run, the user could do something that will cause a conflict

jachro: This is a fair point. To be on the save side it seems reasonable to put that into `init-container`.

jupyterhub/spawners.py
```diff
@@ -350,6 +350,14 @@ def get_pod_manifest(self):
             )
             self.init_containers.append(init_container)
 
+            self.lifecycle_hooks={
+                "preStop": {
+                    "exec": {
+                        "command": ["/bin/sh", "-c", "/usr/local/bin/pre-stop.sh"]
```
rokroskar: This will fail on older images - could we do something like to make it backward-compatible?

`"command": ["/bin/sh", "-c", "/usr/local/bin/pre-stop.sh", "||", "true"]`?

jachro: Good spot! I changed that and tested. All works nice.

PR #135, opened 2019-04-12
Update JupyterHub version

rokroskar:
* Remove `RenkuDockerSpawner`
* Use jupyterhub 0.9.6

Related to https://github.com/SwissDataScienceCenter/renku-jupyter/pull/30


jachro:
I went through the changes, I like all these removals but it's hard for me to say whether it will still work :)

PR #141, opened 2019-04-18
Add notebook server logs endpoint

rokroskar:
The logs for the notebook pod can be fetched from

`<SERVICE_PREFIX>/<namespace>/<project>/<commit-sha>/logs`

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -241,13 +241,17 @@ class RenkuKubeSpawner(SpawnerMixin, KubeSpawner):
         self.extra_annotations = {
             RENKU_ANNOTATION_PREFIX + 'namespace':
                 options.get('namespace'),
-            RENKU_ANNOTATION_PREFIX + 'project':
+            RENKU_ANNOTATION_PREFIX + 'projectName':
                 options.get('project'),
+            RENKU_ANNOTATION_PREFIX + 'projectId':
+                "{}".format(options.get('project_id')),
             RENKU_ANNOTATION_PREFIX + 'branch':
                 options.get('branch'),
             RENKU_ANNOTATION_PREFIX + 'commit-sha':
                 options.get('commit_sha')
         }
+
+        self.delete_grace_period = 30
 
         pod = yield super().get_pod_manifest()
 
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -26,7 +26,7 @@ from urllib.parse import urlsplit, urlunsplit
 import escapism
 from tornado import gen, web
 
-RENKU_ANNOTATION_PREFIX = 'renku.io'
+RENKU_ANNOTATION_PREFIX = 'renku.io/'
 """The prefix for renku-specific pod annotations."""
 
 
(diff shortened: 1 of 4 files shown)


src/notebooks_service.py
```diff
@@ -229,36 +239,54 @@ def ui(path):
...
+    urljoin(SERVICE_PREFIX, '<namespace>/<project>/<commit_sha>/logs'),
+    methods=['GET']
+)
+@authenticated
+def server_logs(user, namespace, project, commit_sha):
+    """"Return the logs of the running server."""
+    server = get_user_server(user, namespace, project, commit_sha)
+    if server:
+        # status = SERVER_STATUS_MAP.get(server.get('pending'), 'not found')
+        return v1.read_namespaced_pod_log(
```
lorenzo-cavazzi: Should we return `JSON` instead of plain text? This may simplify limiting the number of rows, pagination, ...

rokroskar: sure - what should it look like? The k8s api just returns plain text logs, for example

jupyterhub/spawners.py
```diff
@@ -239,20 +239,16 @@ def get_pod_manifest(self):
 
         # add git project-specific annotations
         self.extra_annotations = {
-            RENKU_ANNOTATION_PREFIX + '/namespace':
+            RENKU_ANNOTATION_PREFIX + 'namespace':
                 options.get('namespace'),
-            RENKU_ANNOTATION_PREFIX + '/projectName':
+            RENKU_ANNOTATION_PREFIX + 'project':
```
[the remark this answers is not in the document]
rokroskar: good catch, I believe this is a rebasing mistake (this PR was open long before the one that implemented the change i just overwrote...)

src/notebooks_service.py
lorenzo-cavazzi: This is a log line
```
[D 2019-04-23 08:53:34.263 SingleUserNotebookApp application:555] Looking for jupyter_notebook_config in /home/jovyan/.jupyter
```

* A (possibly) easy solution could be having each line as a string in an array.
```
{
	data: [
		"line 1 plain text",
		"line 2 plain text",
		...
	]
}
```
It's better if we consider a "logical" line, so be careful with multi line logs -- should be easy because it seems they always start with a `tab` instead of a `[`

* A wonderful solution may be having some structured information like
```
{
	data: [
		{
			datetime: "2019-04-23 08:53:34.262",
			source: "application",
			user: "? - SingleUserNotebookApp is not so useful"
			line: 555,
			text: "Looking for jupyter_notebook_config in /home/jovyan/.jupyter"
		},
		{
			...
		},
		...
	]
}
```

lorenzo-cavazzi: When I say "single logical line" I mean something like this
```
[D 2019-04-23 08:53:36.998 SingleUserNotebookApp config_manager:98] Paths used for configuration of jupyter_notebook_config: 
    	/etc/jupyter/jupyter_notebook_config.json
```

rokroskar: Hmm... This seems like maybe a bit overkill - I need to have a look at how e.g. the docker api does this.

rokroskar: We won't always have the same log structure - if, for example rstudio or Matlab are used the logs might be different so we would need to parse them all. One solution would be to read the logs from a logging framework once we deploy one.

lorenzo-cavazzi: Ok then, I suggest to go for the a simple solution now (either keeping plain text logs or having a list of strings) and leave further improvements for later.

rokroskar: yes, I propose we start with just plaintext and move on from there.

ableuler: Returning plain text is fine, but please make sure to set the corresponding `accept` and `content-type` headers in the request and the response respectively.

rokroskar: @ableuler `content-type = text/html` -- should it be something different?

ableuler: is it html? I think `text/plain` would be appropriate here.

jupyterhub/spawners.py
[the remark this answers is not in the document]
rokroskar: fixed

src/notebooks_service.py
rokroskar: done

PR #152, opened 2019-05-09
Add scheme and hostname to notebook server urls

### LINKED ISSUE #95: Return absolute instead of relative URLs
URLs returned by the notebook service api should be absolute, i.e. include the host name. 
```
...
url: "/jupyterhub/user/cramakri/cramakri-weather-2d-9789210/"
```
should become 
```
'''
url: "https://renkulab.io/jupyterhub/user/cramakri/cramakri-weather-2d-9789210/"
```
such that the UI can open that URL directly when opening a notebook server tab for the user.

ableuler:
Closes #95.

---
Merge together with https://github.com/SwissDataScienceCenter/renku-ui/pull/460

### DIFF
--- a/helm-chart/renku-notebooks/templates/deployment.yaml
+++ b/helm-chart/renku-notebooks/templates/deployment.yaml
@@ -51,6 +51,8 @@ spec:
               value: {{ .Values.jupyterhub.hub.baseUrl }}services/notebooks/
             - name: JUPYTERHUB_BASE_URL
               value: {{ .Values.jupyterhub.hub.baseUrl }}
+            - name: JUPYTERHUB_ORIGIN
+              value: {{ template "notebooks.http" . }}://{{ .Values.global.renku.domain }}
             - name: JUPYTERHUB_CLIENT_ID
               value: {{ .Values.jupyterhub.hub.services.notebooks.oauth_client_id }}
             {{ if .Values.gitlab.registry.secret }}
--- a/src/notebooks_service.py
+++ b/src/notebooks_service.py
@@ -36,6 +36,9 @@ from flask import (
 )
 from jupyterhub.services.auth import HubOAuth
 
+JUPYTERHUB_ORIGIN = os.environ['JUPYTERHUB_ORIGIN']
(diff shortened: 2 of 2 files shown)


rokroskar:
added it to next minor-revision milestone as it breaks UI compatibility

helm-chart/renku-notebooks/templates/deployment.yaml
```diff
@@ -51,6 +51,8 @@ spec:
               value: {{ .Values.jupyterhub.hub.baseUrl }}services/notebooks/
             - name: JUPYTERHUB_BASE_URL
               value: {{ .Values.jupyterhub.hub.baseUrl }}
+            - name: JUPYTERHUB_ORIGIN
+              value: {{ template "notebooks.http" . }}://{{ .Values.global.renku.domain }}
```
rokroskar: should the default be `.Values.jupyterhub.url`?

ableuler: In most of our charts it's the other way around: a global chart value is the default which can be overridden by a local one. However, I would prefer not to do this here. In fact I don't see a need for this as a standard pattern and I would like to move away from it. It just results in even more places where the same deployment parameter can be set. Either a value is the same across all renku services and should thus be global, or it's something that should be set for each service specifically and thus local. Or am I missing something here?

rokroskar: ok but the default values file for the notebooks service doesn't define `renku.domain` in its `global` section

ableuler: But in this case the `JUPYTERHUB_API_URL` just a few lines above will also be undefined. I tried get away with minimal chart changes for this one and use the values which are already there. I agree there's more to be fixed, but let's try to keep PRs small.

rokroskar:
@ableuler we can now merge these PRs...

ableuler:
> @ableuler we can now merge these PRs...

👍 I will probably have to do some slightly more involved rebase for the one on the UI side...

rokroskar:
ok so do that first and then we merge this one

PR #156, opened 2019-05-17
137 use configmap

mohammad-sdsc:


helm-chart/renku-notebooks/templates/configmap.yaml
```diff
@@ -23,3 +23,31 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   SENTRY_DSN: {{ .Values.jupyterhub.singleuser.sentryDsn }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hub-helper-scripts
```
rokroskar: this is a very minor thing, but technically this is a helper script for jupyter notebooks not the hub... maybe the naming will be misleading?

mohammad-sdsc: `notebook-helper-scripts` then?

rokroskar: 👍

helm-chart/renku-notebooks/values.yaml
```diff
@@ -143,7 +143,13 @@ jupyterhub:
       name: renku/singleuser
       tag: latest
     storage:
-      type: none
```
rokroskar: don't remove this!

helm-chart/renku-notebooks/values.yaml
```diff
@@ -144,6 +144,15 @@ jupyterhub:
...
     storage:
       type: none
+      extraVolumes:
+        - name: notebook-helper-scripts-volume
+          configMap:
+            name: notebook-helper-scripts
+            defaultMode: 0755
+      extraVolumeMounts:
+        - name: notebook-helper-scripts-volume
+          mountPath: /usr/local/bin/pre-stop.sh
```
rokroskar: I guess this works for a single script, but what if we had multiple?

rokroskar: ah nevermind, it's fine 👍

PR #160, opened 2019-05-23
enhancement: set ephemeral storage limit on pods

mohammad-sdsc:


helm-chart/renku-notebooks/values.yaml
```diff
@@ -155,6 +155,11 @@ jupyterhub:
           subPath: pre-stop.sh
     defaultUrl: /lab
     sentryDsn: ''
+    extraResource:
```
rokroskar: We don't want to set defaults here - these values should be commented out and provided as documentation since the exact value will vary from deployment to deployment.

mohammad-sdsc: So, I leave them here but comment them out or I need to do something somewhere else?

rokroskar: yes, we'll provide these as documentation and then implement the change in our deployments

Issue #138, opened 2019-04-15
Push `renku/isolation` branches for preStop


jirikuncar:
We should consider pushing all `renku/*` branches - most importantly `renku/isolation/*` ones (at least the unmerged ones).

addresses #130

PR #164, opened 2019-05-28
feat: push unpushed changes to origin

mohammad-sdsc:


rokroskar:
I'm not entirely sure this is the solution - the problem is that maybe the person isn't ready to push to origin because of known conflicts etc. The only way to be sure is to basically push this also to an autosave branch IMO.

mohammad-sdsc:
I see. I thought that there is only one branch moving forward and did not know that there might be possible conflicts.
The autosave branch is already pushing un-pushed changes to its own remote since it is branching from the top. Isn't this enough then?

rokroskar:
ah wait - you're right! Of course... 

there is this related issue: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/138

The "isolation" branches are there for renku executions that run in parallel. Currently we are not doing anything to save those.

rokroskar:
sorry this isn't quite done. In the case where I commit some changes and my notebook server gets terminated, the unpushed changes don't go to the autosave branch because of https://github.com/SwissDataScienceCenter/renku-notebooks/blob/597b3ba19cf6fe6280a2a681d46e589d3b13a89a/helm-chart/renku-notebooks/templates/configmap.yaml#L43-L45

I think we should change this behavior to check if the current `HEAD` is different than the value of `CI_GIT_COMMIT_SHA` environment variable. If not, then we exit, if yes then we push to the autosave branch.

PR #168, opened 2019-06-03
fix code style

rokroskar:
* adds `flake8` configuration
* switch to `black` from `yapf`
* run pytest on travis
* install `pre-commit` hooks for `black` and `flake8`


.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 88
+ignore = E203, W503
```
ableuler: For VS code, go to `code > preferences > settings` and search for `Insert Final Newline`

renku_notebooks/__init__.py
```diff
@@ -38,16 +36,16 @@ def __init__(self, app):
         self.app = app
 
     def __call__(self, environ, start_response):
-        script_name = environ.get('HTTP_X_SCRIPT_NAME', '')
+        script_name = environ.get("HTTP_X_SCRIPT_NAME", "")
```
ableuler: I'd prefer sticking to single quotes as they have been used in the project so far (also `renku-python` and `renku-gateway` use single quotes).

.flake8
rokroskar: oh I thought I had that on... maybe it was only local to a project and not the global config.

renku_notebooks/__init__.py
rokroskar: `black` doesn't care what you think! :)

rokroskar: here is a nice discussion on the topic: https://github.com/python/black/issues/118

.flake8
rokroskar: done

renku_notebooks/__init__.py
rokroskar: I'm not going to be religious about quotes. After spending way too much time configuring `yapf` options, not having a single option to worry about and getting decent-looking code out of it is a win-win for me.

ableuler: Ok, I wasn't aware that `black` was such an opinionated project and I'm a bit astonished that it allows you to configure the line-length but not if you prefer single or double quotes... But fine for me.

rokroskar: it allows you to ignore its preference for double-quotes if you wish... but then you end up with inconsistent quotes.

PR #170, opened 2019-06-06
refactor(charts): move registry secret from renku repo

### LINKED ISSUE #167: make the registry secret in the notebooks chart
The secret used for pulling private images is currently created in the main renku chart - we should move it to the notebooks service chart instead.

rokroskar:
Note that this change allows non-root admin tokens to be used. This should make it easier to rotate them periodically. 

```
gitlab:
  registry:
    username: a_user
    token: 12345
```

closes #167

### DIFF
--- a/helm-chart/renku-notebooks/templates/registry-secret.yaml
+++ b/helm-chart/renku-notebooks/templates/registry-secret.yaml
@@ -1,0 +1,15 @@
+{{- $auth := printf "%s:%s" .Values.gitlab.registry.username .Values.gitlab.registry.token | b64enc -}}
+{{- $registry := .Values.gitlab.registry.host | replace "http://" "" | replace "https://" "" | replace ":443" "" -}}
+{{- $secret := printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"root@example.com\",\"auth\":\"%s\"}}}" $registry .Values.gitlab.registry.username .Values.gitlab.registry.token $auth -}}
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: {{ template "notebooks.fullname" . }}-registry
+  labels:
+    app: {{ template "notebooks.name" . }}
+    chart: {{ template "notebooks.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  .dockerconfigjson: {{ $secret | b64enc | quote }}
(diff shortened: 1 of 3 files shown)


helm-chart/renku-notebooks/templates/registry-secret.yaml
```diff
@@ -0,0 +1,15 @@
+{{- $auth := printf "%s:%s" .Values.gitlab.registry.username .Values.gitlab.registry.token | b64enc -}}
+{{- $registry := .Values.gitlab.registry.host | replace "http://" "" | replace "https://" "" | replace ":443" "" -}}
+{{- $secret := printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"root@example.com\",\"auth\":\"%s\"}}}" $registry .Values.gitlab.registry.username .Values.gitlab.registry.token $auth -}}
```
ableuler: Does that email have to be here? It could be any admin, no?

helm-chart/renku-notebooks/values.yaml
```diff
@@ -40,26 +42,30 @@ securityContext:
 # turns on flask debug mode
 debug: true
 
+## Default server options - these will be provided to the user in the UI
```
ableuler: I think this should be in a separate PR.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -178,16 +177,30 @@ jupyterhub:
...
       ## use openssl rand -hex 32 to generate clientSecret
       clientSecret:
-      callbackUrl: '""' # Forces the default callback url
+      callbackUrl: # Leaving it null forces the default callback url
   proxy:
     service:
       type: ClusterIP
     https:
       enabled: false
+  ## Enable idle user server culling
```
ableuler: Also material for a separate PR.

helm-chart/renku-notebooks/templates/registry-secret.yaml
rokroskar: true I hadn't noticed that. It doesn't actually seem to match the email in gitlab for the root user so it doesn't look like it's being used. I would leave it for now.

helm-chart/renku-notebooks/values.yaml
rokroskar: hmm... just adding a few lines of documentation to the values file... but ok.

ableuler: Sorry, I thought that had ended up in here by accident because it has nothing to do with the title/description of the PR or the commit message. I'm perfectly fine with having "non-atomic" PRs, but at least having two commits for things that have nothing in common would indicate that this was intended.

rokroskar: no worries, I already have a new PR for that, can include more documentation (and fix the README) there as well

rokroskar:
I tested this on my dev environment using my own token and launching a project from one of my private repos - seems to work as intended.

PR #174, opened 2019-06-12
140 push to remote when autosaving

mohammad-sdsc:


git-clone/entrypoint.sh
```diff
@@ -32,7 +32,6 @@ done
 if [ -z "$AUTOSAVE_REMOTE_BRANCH" ] ; then
   (git checkout ${BRANCH} || git checkout -b ${BRANCH})
   git submodule init && git submodule update
-  git reset --hard $COMMIT_SHA
```
mohammad-sdsc: I am not sure what this line does. However, it creates a bug:

1. Start a server in UI and connect to it.
2. In its command line, make some changes, commit, and push.
3. Stop the server.
4. Start the server again and go to its command line.
5. `git log` shows that your current `HEAD` points at the previous commit while remote still points to the commit pushed in step 2.

mohammad-sdsc:
This was tested with all combinations of `un-pushed` and `un-saved` changes.

git-clone/entrypoint.sh
mohammad-sdsc: Am I doing something wrong/not-supposed-to-do here?

rokroskar: no, what is happening is that the UI state is still pointing to the previous commit - so when you restart the server it starts it for that commit and not the latest one. If you refresh the UI, you should get the correct behavior. This is being addressed in the UI.

rokroskar: so this line is actually needed to be sure that we are on the correct commit.

mohammad-sdsc: Uncommitted!

PR #184, opened 2019-06-19
feat: spawn new server for different branch

### LINKED ISSUE #90: notebooks: launch multiple servers for different branches
- [ ] ` ?branch=<name>` should create a server with different name that includes branch name

https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/src/notebooks_service.py#L102-L113

rokroskar:
Instead of putting the project name and commit sha into the server name, we form the servername by hashing the string 

```
{namespace}{project}{commit-sha}{ref}
```

where `ref` for the time being is assumed to be a branch but could be a tag, for instance.

The resulting pod name is something like `jupyter-rokroskar-9ecaa0eeed735403` (16-digit hash) and the server name is just the hash, i.e. in this case `rokroskar/9ecaa0eeed735403`. 

---

closes #90

### DIFF
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -79,6 +79,7 @@ def notebook_status(user, namespace, project, commit_sha, notebook=None):
 @authenticated
 def launch_notebook(user, namespace, project, commit_sha, notebook=None):
     """Launch user server with a given name."""
+    branch = request.args.get("branch", "master")
     # 0. check if server already exists and if so return it
     name = server_name(namespace, project, commit_sha)
     server = get_user_server(user, namespace, project, commit_sha)
@@ -124,7 +125,7 @@ def launch_notebook(user, namespace, project, commit_sha, notebook=None):
         )
 
     payload = {
-        "branch": request.args.get("branch", "master"),
+        "branch": branch,
         "commit_sha": commit_sha,
         "namespace": namespace,
         "notebook": notebook,
--- a/renku_notebooks/util/jupyterhub_.py
+++ b/renku_notebooks/util/jupyterhub_.py
@@ -18,11 +18,9 @@
 """Functions for interfacing with JupyterHub."""
 
 import os
-import string
-from functools import partial
+from hashlib import md5
 from urllib.parse import urljoin
 
-import escapism
 from flask import Blueprint, current_app
 
 from .. import config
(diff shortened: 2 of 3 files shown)


renku_notebooks/api/notebooks.py
```diff
@@ -79,6 +79,7 @@ def notebook_status(user, namespace, project, commit_sha, notebook=None):
 @authenticated
 def launch_notebook(user, namespace, project, commit_sha, notebook=None):
     """Launch user server with a given name."""
+    branch = request.args.get("branch", "master")
```
lorenzo-cavazzi: shouldn't the route at line 77/78 be updated to something like `"<namespace>/<project>/<branch>/<commit_sha>"` to keep things coherent?

rokroskar: A branch is not a unique property while a commit is (e.g. you can have two branches on the exact same commit). That's why it was excluded to begin with. What about passing the commit *and* the branch as data? E.g. you would make a post to `<namespace>/<project>` and specify a json with 

```
{
  "commit-sha": "12345",
  "ref": "some-branch"
}
```

rokroskar: The `commit-sha` would be required and `ref` optional.

mohammad-sdsc: Everywhere in the notebook service with use the combination of `<namespace>/<project>/<commit_sha>` to uniquely identify an endpoint. If we spawn servers from different branches then we need `<branch>` to be able to do that (e.g. two different servers from two different branches on the same commit should be identifiable for DELETE). So I prefer to have it in the endpoint url as @lorenzo-cavazzi suggested.

mohammad-sdsc: Moreover, I prefer to stick to `branch` and not `ref` to avoid confusion.

mohammad-sdsc: Also, if you don't mind I'll include your changes in my PR and update the notebook services API as well.

rokroskar: but then it should be `<namespace>/<project>/<commit-sha>/<branch>` since you can have multiple branches per commit. 

What should a `GET` return? I suppose on `<namespace>/<project>` it should return a list of servers running per commit; on `<namespace>/<project>/<commit-sha>` it should return the list of servers identified by branch for that commit, and on `<namespace>/<project>/<commit-sha>/<branch>` it should return the status of the uniquely-identified server.

rokroskar: > Also, if you don't mind I'll include your changes in my PR and update the notebook services API as well.

why? this PR has nothing to do with refactoring the other endpoint.

rokroskar: from a rest api point of view, I don't really like the URL sprawl - a POST on the `<namespace>/<project>/<commit>` endpoint doesn't make sense because we cannot create any objects there. The object we are creating with a POST is a server for the project. That's why I like the idea of having a single endpoint be `<namespace>/<project>` (or with `/servers`) at the end - a GET retrieves the list of running servers, a POST creates a new one. A GET with `commit` and `branch` in the payload gets you the specific server.

rokroskar:
Hmm true I forgot about the failing tests - it's because of the change to the server name.

rokroskar:
it was a trivial change - should be fixed now. Please re-approve if the tests pass.

lorenzo-cavazzi:
I found an issue with this -- I am reporting here because #186 will _probably_ solve this automatically.

When a server is started for the combination `(commit1,branch1)`, it's not possible to start another server for the combination `(commit1,branch2)`. Could this be somehow related to the `POST` url without the explicit branch and some missing checks? `/.../commit1?branch=branch2` instead of something like ``/.../commit1/branch2``

mohammad-sdsc:
That's expected and that'a what's I meant when I said API would be in an inconsistent state. When creating new notebook, API first checks if there is another one which is already running and returns that one. This check cannot differentiate between different branches because we only implemented that feature in POST requests and not for GET.
You are right, https://github.com/SwissDataScienceCenter/renku-notebooks/issues/186 will solve this issue.

PR #198, opened 2019-08-13
use a system-agnostic workdir

rokroskar:
This PR modifies the location of the attached volume that contains the cloned project to `/work/<project-name>`.


rokroskar:
Thanks for catching that @Panaetius - I removed that config since it's set explicitly in the spawners code.

PR #203, opened 2019-08-21
fix annotation bug

rokroskar:
Pod annotations pass through the same code that makes pod names dns-safe - this means that we were shortening them in a way that made them unusable. ~~This PR fixes the situation by only running pod and pvc templates through the expansion code/filter.~~ This PR fixes the situation by removing the custom `expand_user_properties` code and instead just handling the pod name length individually.

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -145,6 +145,11 @@ class RenkuKubeSpawner(SpawnerMixin, KubeSpawner):
         repository = yield self.git_repository()
         options = self.user_options
 
+        # make sure the pod name is less than 64 characters - if longer, keep
+        # the last 16 untouched since it is the server hash
+        if len(self.pod_name) > 63:
+            self.pod_name = self.pod_name[:47] + self.pod_name[-16:]
+
         # Process the requested server options
         server_options = options.get("server_options", {})
         self.default_url = server_options.get("defaultUrl")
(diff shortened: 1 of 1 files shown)


jupyterhub/spawners.py
```diff
@@ -270,8 +270,10 @@ def _expand_user_properties(self, template):
...
         the server name hash.
 
-        Code adapted from
-        https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py
+        If the string being processed is not the pod or pvc template,
+        let it pass through.
         """
-        rendered = super()._expand_user_properties(template)
-        return rendered if len(rendered) <= 63 else rendered[:47] + rendered[-16:]
+        if template == self.pod_name_template or template == self.pvc_name_template:
```
mohammad-sdsc: Shouldn't we call the base class method anyway and only doing the shortening if this condition holds:
```
rendered = super()._expand_user_properties(template)
if template == self.pod_name_template or template == self.pvc_name_template:
    return rendered if len(rendered) <= 63 else rendered[:47] + rendered[-16:]
return rendered
```

rokroskar: hmm maybe - but annotations, for example, don't need to be escaped at all and the jupyterhub code is passing them through this function. That means that you might get a different annotation than you expect (I will make an issue/PR to kubespawner about this). What's weird is that they pass the annotation values through `expand_user_properties` but it's the keys in annotations that have character and length limitations, not the values.

rokroskar: On second thought I think you're right, since the rest of the code expects that behavior we should leave it for now. I will change it.

rokroskar:
@mohammad-sdsc I realized we no longer needed to override the `expand_user_properties` method - I think this is much much simpler now.

mohammad-sdsc:
Cool! I never liked that part of the code.

rokroskar:
> I never liked that part of the code.

it seemed necessary at some point... /shrug

PR #204, opened 2019-08-22
chore: bump version and changelog

rokroskar:


rokroskar:
> However I dont see the 0.5.2 version deployed on rok.dev.renku.ch yet.
yes this is a bit of a hack because of the way chartpress tags versions - the version in our `Chart.yaml` is actually the upcoming version. The released one is one semantic version lower.

PR #211, opened 2019-09-12
feat: use project name in the notebook URL

lorenzo-cavazzi:
Server name created by JupyterHub includes up to 46 chars from the reference project name. This is a possible quick solution to SwissDataScienceCenter/renku-ui#584 *apparently* without any significative side effect.

Preview available here: https://lorenzotest.dev.renku.ch

Relevant changes: the server name is different, so the JupyterLab url will also be affected accordingly
Here is the url comparison of a project named `zurich-bikes-tutorial`

Screenshot from 2019-09-11 16-37-56

fix SwissDataScienceCenter/renku-ui#584

### DIFF
--- a/renku_notebooks/util/jupyterhub_.py
+++ b/renku_notebooks/util/jupyterhub_.py
@@ -37,7 +37,7 @@ def make_server_name(namespace, project, branch, commit_sha):
     """Form a 16-digit hash server ID."""
     server_string = f"{namespace}{project}{branch}{commit_sha}"
     return "{project}-{hash}".format(
-        project=project[:46], hash=md5(server_string.encode()).hexdigest()[:16]
+        project=project[:54], hash=md5(server_string.encode()).hexdigest()[:8]
     )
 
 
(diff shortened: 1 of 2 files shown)


renku_notebooks/util/jupyterhub_.py
```diff
@@ -37,18 +36,9 @@
...
-    return md5(server_string.encode()).hexdigest()[:16]
-
-
-def notebook_url(user, server_name, notebook=None):
-    """Form the notebook server URL."""
-    notebook_url = urljoin(
-        os.environ.get("JUPYTERHUB_BASE_URL"),
-        "user/{user[name]}/{server_name}/".format(user=user, server_name=server_name),
+    return "{project}-{hash}".format(
+        project=project[:46], hash=md5(server_string.encode()).hexdigest()[:16]
```
rokroskar: we could even probably use only a 7-character hash and it would be more than enough - what do you think?

lorenzo-cavazzi: right, I guess we can safely go down to 7 or 8 chars hash.
Not sure if going from the first 46 chars to 54 would be a noticeable difference for the user :slightly_smiling_face:

PR #213, opened 2019-09-17
style: improve label and ordering of server configuration options

ciyer:


helm-chart/renku-notebooks/values.yaml
```diff
@@ -48,27 +48,33 @@ debug: true
...
+  #   displayName: Number of GPUs
+  #   type: int
+  #   default: 0
+  #   range: [0, 0]
   gpu_request:
-    displayName: GPU request
-    type: int
-    default: 0
-    range: [0, 0]
-  defaultUrl:
```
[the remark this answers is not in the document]
rokroskar: ah nevermind, I see you moved it up

helm-chart/renku-notebooks/values.yaml
```diff
@@ -48,27 +48,33 @@ debug: true
 ## Note that requests are also limits, i.e. a user's jupyter kernel
 ## that exceeds the requested memory limit will be terminated/restarted.
 serverOptions:
+  defaultUrl:
+    displayName: Default Environment
+    type: enum
+    default: /lab
+    ## add /rstudio to the list for R images
```
rokroskar:
```suggestion
    ## add /rstudio to the list for R projects
```

helm-chart/renku-notebooks/values.yaml
```diff
@@ -48,27 +48,33 @@ debug: true
...
     type: enum
     default: 1G
     options: [1G, 2G]
+  # Int values are also supported
+  # gpu_request:
+  #   displayName: Number of GPUs
+  #   type: int
+  #   default: 0
+  #   range: [0, 0]
   gpu_request:
```
lorenzo-cavazzi: We could comment this out (and remove the already commented `gpu_request` above) and leave the setting in the dev python script, otherwise there is no clean way for the admin to remove the option completely. It may be desirable in environments where no GPUs are available at all, so the user can't even ask the admin to provide some.

ciyer: I think this should be here even if the user cannot make a choice because we need to inform the user about what resources they will get when they start an environment.

ciyer: After discussing with Rok, we have decided to leave in the option, even if it is 0.

helm-chart/renku-notebooks/values.yaml
ciyer: Good suggestion. Done.

PR #214, opened 2019-09-17
chore: move comment to avoid chartpress problem

lorenzo-cavazzi:
This *apparently* useless PR addresses a silly issue I experience with `chartpress`.
In short: every time I run `chartpress` it duplicates/moves data for the `callbackUrl` value (see screen below). I guess it doesn't like a comment next to an empty value :man_shrugging: 

Screenshot from 2019-09-17 16-08-35

### DIFF
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -171,7 +171,8 @@ jupyterhub:
       clientId: jupyterhub
       ## use openssl rand -hex 32 to generate clientSecret
       clientSecret:
-      callbackUrl: # Leaving it null forces the default callback url
+      # Leaving it null forces the default callback url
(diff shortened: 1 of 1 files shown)


rokroskar:
we've run into this before - I thought it was a `ruamel.yaml` version problem - it could also be that you need more than one space between the `:` and the `#` but in any case, putting it on a different line definitely fixes the issue :)

rokroskar:
I just ran `chartpress` and don't see this - which version of `ruamel.yaml` are you using?

lorenzo-cavazzi:
I was on `0.15.51`. Last time I only tried to update `chartpress` and I already had the last version, but now I have just tried to bump `ruamel` and it's working :tada: (the version is `0.16.5`).
I guess we can close the issue :slightly_smiling_face:

PR #215, opened 2019-10-01
WIP feat: update jupyterhub to 1.0.0

### LINKED ISSUE #183: Upgrade JupyterHub to version 1.x
At the moment, we are using JupyterHub version 0.9.6. An upgrade to the latest version might solve some corner cases that we see but also might break some workarounds that we put in place to make JupyterHub work.

lorenzo-cavazzi:
This is an attempt to upgrade JupyterHub version to 1.0.0

It seems to solve an annoying problem with hanging pods (SwissDataScienceCenter/renku-ui#361) that has been bothering us for a while. Basically, the eviction policy now works correctly on pending pods after the timeout (here reduced to 5 minutes).

This update seems to add a new security feature that pops up an "Authorize acces" page to the user when he logs in. We probably want to get rid of this.

Preview available here: https://lorenzotest.dev.renku.ch

image

P.S. mind that this update requires to upgrade the DB https://zero-to-jupyterhub.readthedocs.io/en/latest/upgrading.html#database

fix #183


helm-chart/renku-notebooks/values.yaml
```diff
@@ -126,7 +129,7 @@ jupyterhub:
         c.Spawner.debug = bool(os.getenv('DEBUG', False))
 
         # Increase pod initialization timeout to 15 minutes
```
ableuler:
```suggestion
        # Increase pod initialization timeout to 5 minutes
```

jupyterhub/requirements.txt
```diff
@@ -17,6 +17,6 @@
 # limitations under the License.
 
 # Installing from Git until a new version of kubespawner (introducing 'delete_grace_period') is released.
--e git://github.com/jupyterhub/kubespawner.git@46a4b109c5e657a4c3d5bfa8ea4731ec6564ea13#egg=jupyterhub-kubespawner
-jupyterhub==0.9.6
+-e git://github.com/jupyterhub/kubespawner.git@01373369044f8fd9754956bb6a3b3cfeecffb939#egg=jupyterhub-kubespawner
+jupyterhub==1.0.0
```
ableuler: Can't you simply get rid of this line? Doesn't any `jupyterhub/k8s-hub:0.9-dev` image come with jupyterhub 1.0.0 installed?

helm-chart/renku-notebooks/values.yaml
lorenzo-cavazzi: ops! thanks for spotting this

jupyterhub/requirements.txt
lorenzo-cavazzi: that's true, I can remove it. I am wondering if it makes sense to leave it here anyway to make it clearer that `jupyterhub` must be updated also here -- otherwise it's not that intuitive for someone looking for `jupyterhub` that `jupyterhub/k8s-hub:0.9-dev` requires `jupyterhub` 1.0.0

lorenzo-cavazzi:
@ableuler after checking the documentation thoroughly (and trying a couple of solutions) without identifying how to hide this message, I asked in the gitter chat and they answered me they are already working on this feature. The idea is to provide a `whitelist` variable where you can list all the services automatically authorized without requiring explicit confirmation.

I would say we can wait to upgrade until this is available (I'll add a reference to the relevant issue on the JupyterHub repo as soon as I identify it -- they told me there is one already but I couldn't find it)

lorenzo-cavazzi:
It seems that the `authorize access` message con be easily removed in a future JH release thanks to this jupyterhub/jupyterhub#2767 
I'll get back to this problem soon -- maybe it worth to wait until the version `1.1` is released jupyterhub/jupyterhub#2830

rokroskar:
great news, thanks for keeping up with it @lorenzo-cavazzi!

lorenzo-cavazzi:
version `1.1.0` is released, so we may superseed this PR with a trial on that one. The option to remove that extra confirmation message should be `Service.oauth_no_confirm`
REF: https://jupyterhub.readthedocs.io/en/stable/changelog.html#id1

lorenzo-cavazzi:
I close this in favor of another PR going to introduce JH 1.1.0

PR #216, opened 2019-10-11
feat: gracefully handle predictable k8s api errors when querying for logs

lorenzo-cavazzi:
During the spawning phase, querying kubernetes logs API may result in a predictable error. This PR returns a plaintext message instead of a `400` error response

### DIFF
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -216,15 +216,15 @@ def server_options(user):
 
 @bp.route("logs/<server_name>")
 @authenticated
-def server_logs(user, server_name):
-    """"Return the logs of the running server."""
+def server_logs(user, server_name, max_log_lines=100):
+    """Return the logs of the running server."""
     if request.environ["HTTP_ACCEPT"] == "application/json":
         return make_response("Only supporting text/plain.", 406)
     server = get_user_server(user, server_name)
     if server:
         pod_name = server.get("state", {}).get("pod_name", "")
         try:
-            logs = read_namespaced_pod_log(pod_name)
+            logs = read_namespaced_pod_log(pod_name, max_log_lines)
         # catch predictable k8s api errors and return a significative string
         except ApiException as e:
             logs = ""
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -222,7 +222,13 @@ def server_logs(user, server_name):
     server = get_user_server(user, server_name)
     if server:
         pod_name = server.get("state", {}).get("pod_name", "")
-        logs = read_namespaced_pod_log(pod_name)
+        try:
+            logs = read_namespaced_pod_log(pod_name)
+        # catch predictable k8s api errors and return a significative string
+        except Exception as e:
+            if hasattr(e, "body"):
+                k8s_error = json.loads(e.body)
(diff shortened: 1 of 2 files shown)


renku_notebooks/api/notebooks.py
```diff
@@ -222,7 +222,13 @@ def server_logs(user, server_name):
     server = get_user_server(user, server_name)
     if server:
         pod_name = server.get("state", {}).get("pod_name", "")
-        logs = read_namespaced_pod_log(pod_name)
+        try:
+            logs = read_namespaced_pod_log(pod_name)
+        # catch predictable k8s api errors and return a significative string
+        except Exception as e:
```
mohammad-sdsc: Can you put a more specific Exception type here?

renku_notebooks/api/notebooks.py
```diff
@@ -222,7 +222,13 @@ def server_logs(user, server_name):
...
+            if hasattr(e, "body"):
+                k8s_error = json.loads(e.body)
+                logs = f"Logs unavailable: {k8s_error['message']}"
```
mohammad-sdsc: You should set a default value for `logs` before the if statement; otherwise, you'll get an undefined name error when the if condition is false:
```
logs = ""
if hasattr ...
```

renku_notebooks/api/notebooks.py
lorenzo-cavazzi: sure, switched to `ApiException`

renku_notebooks/api/notebooks.py
lorenzo-cavazzi: good point! I initialized the variable as an empty string as suggested

renku_notebooks/util/kubernetes_.py
```diff
@@ -183,4 +183,4 @@ def get_server_url(pod):
 
 
 def read_namespaced_pod_log(pod_name):
-    return v1.read_namespaced_pod_log(pod_name, kubernetes_namespace)
+    return v1.read_namespaced_pod_log(pod_name, kubernetes_namespace, tail_lines=100)
```
mohammad-sdsc: What is this 100? Is this the maximum number of lines that you need in UI? Would you please rename it to something more expressive and use the named-constant here (e.g. `MAX_REQUIRED_LINES = 100`)?

lorenzo-cavazzi: Yes, `tail_lines` is the maximum number of lines (from the most recent) that the function will return. We don't want to flood the UI with useless logs.

If you prefer a more explicit naming we may introduce it as an optional parameter like this

```
def read_namespaced_pod_log(pod_name, max_log_lines = 100):
    return v1.read_namespaced_pod_log(pod_name, kubernetes_namespace, tail_lines = max_log_lines)
```

mohammad-sdsc: Both options are fine with me since we don't pass this value from REST API to this function yet.

lorenzo-cavazzi: I am wondering if it makes sense to add the value directly to the rest API. We won't use this now, but at least we provide the possibility

mohammad-sdsc: I know. That's why I'm saying we can have this either as a named-constant or a parameter. Since we don't pass the parameter, having it as a named-constant is OK as well.

PR #218, opened 2019-11-06
chore: bump chart version

rokroskar:


helm-chart/renku-notebooks/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Notebooks service
 name: renku-notebooks
-version: 0.5.2
+version: 0.0.1-set.by.chartpress
```
lorenzo-cavazzi: with this change don't we always get `0.0.1-<8-digit-commit-sha>`. Do we want this? :thinking:

rokroskar: no definitely not, I had a wrong chartpress version installed

helm-chart/renku-notebooks/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Notebooks service
 name: renku-notebooks
-version: 0.5.2
+version: 0.6.1
```
lorenzo-cavazzi: I was about to write that the version here should be `0.6.0` so that the release will be properly tagged, but I remeber we have already had this conversation multiple times :smile:

rokroskar: yes, though it's been a while so I was expecting this comment :)

rokroskar:
? it failed...

PR #220, opened 2019-11-14
fix: reference environment starting commit on autosave branches

### LINKED ISSUE #219: wrong commit reference in autosaved branch
Currently, when a user works on an interactive environment and it gets stopped without saving, an autosaved branch is created with the following name `renku/autosave/<user>/<branch>/<starting_commit>/<final_commit>`.

The `starting_commit` is the current latest commit on remote and this may lead to problems when a user starts from an older commit and leave unchanged code because it's saved under the wrong commit.

E.G. assume the latest commit is `2222222` and the previous one is `1111111`. Starting an environment from `1111111` will result in `renku/autosave/<user>/<branch>/<2222222>/<final_commit>` instead of `renku/autosave/<user>/<branch>/<1111111>/<final_commit>`

lorenzo-cavazzi:
Use the environment starting commit to compose the autosaved branches.

A preview is available on https://lorenzotest.dev.renku.ch and can easily be tested from the UI

#### How to test
From any project with >= 2 commits (e.g. fork https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial/environments) start a new environment from an old commit, do any work (e.g. `touch aaa`) and stop the notebook without further actions. Trying to start a new environment from the same commit will result in the autosave message, while trying to start from the latest commit won't display any message. The behavior is currently inverted on https://dev.renku.ch/

fix #219

### DIFF
--- a/helm-chart/renku-notebooks/templates/configmap.yaml
+++ b/helm-chart/renku-notebooks/templates/configmap.yaml
@@ -47,16 +47,12 @@ data:
 
     CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
     LOCAL_SHA=`git rev-parse --short HEAD`
-    REMOTE_SHA=`git rev-parse --short @{upstream}`
+    INITIAL_SHA="${CI_COMMIT_SHA:0:7}"
 
-    if [ -z "$REMOTE_SHA" ]; then
-      REMOTE_SHA="${LOCAL_SHA}"
-    fi
-
-    AUTOSAVE_BRANCH="renku/autosave/$JUPYTERHUB_USER/${CURRENT_BRANCH}/${REMOTE_SHA}/${LOCAL_SHA}"
+    AUTOSAVE_BRANCH="renku/autosave/$JUPYTERHUB_USER/${CURRENT_BRANCH}/${INITIAL_SHA}/${LOCAL_SHA}"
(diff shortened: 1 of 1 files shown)


helm-chart/renku-notebooks/templates/configmap.yaml
```diff
@@ -47,16 +47,12 @@ data:
 
     CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
     LOCAL_SHA=`git rev-parse --short HEAD`
-    REMOTE_SHA=`git rev-parse --short @{upstream}`
+    INITIAL_SHA="${CI_COMMIT_SHA:0:7}"
 
-    if [ -z "$REMOTE_SHA" ]; then
-      REMOTE_SHA="${LOCAL_SHA}"
-    fi
```
mohammad-sdsc: Should we keep check (for `$INITIAL_SHA`) or there is a guarantee that `$CI_COMMIT_SHA` will always have the right value?

lorenzo-cavazzi: I believe the `$CI_` variable are always there, those are the values inherited from the pod annotations. If not, the environments won't probably work correctly anyway :slightly_smiling_face:

PR #225, opened 2020-01-15
feat: update server_logs api

### LINKED ISSUE #224: Change API response from text to application/json
There is one API providing `text/plain` response instead of `application/json` (`server_logs` in /renku_notebooks/api/notebooks.py). This should be changed to allow content-based redirection SwissDataScienceCenter/renku-gateway#174
It requires a small fix on the UI as well.

lorenzo-cavazzi:
* return JSON data instead of TEXT
* change the default lines to 250

fix #224

### DIFF
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -216,10 +216,8 @@ def server_options(user):
 
 @bp.route("logs/<server_name>")
 @authenticated
-def server_logs(user, server_name, max_log_lines=100):
+def server_logs(user, server_name, max_log_lines=250):
     """Return the logs of the running server."""
-    if request.environ["HTTP_ACCEPT"] == "application/json":
-        return make_response("Only supporting text/plain.", 406)
     server = get_user_server(user, server_name)
     if server:
         pod_name = server.get("state", {}).get("pod_name", "")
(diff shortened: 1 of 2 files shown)


rokroskar:
I wonder if there is a setting for the notebook server to actually return JSON logs? Here we would just be turning each line into a JSON payload I guess with no added data granularity.

lorenzo-cavazzi:
That would be great but I couldn't find anything about that in the kubernetes library docs  https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#read_namespaced_pod_log

rokroskar:
no, the application itself (the jupyter notebook server in this case) would need to produce a json log.

lorenzo-cavazzi:
I guess it's theoretically possible (taking inspiration from here https://docs.python.org/3/howto/logging-cookbook.html#implementing-structured-logging), but I wonder if it gives us any advantages since it's meant to be read by users and not stored/processed anywhere.

rokroskar:
Well, most likely we will be monitoring those logs via elasticsearch - that's why the json logs would be useful. It's an open issue: https://github.com/jupyterhub/jupyterhub/issues/2368

lorenzo-cavazzi:
I see. That issue doesn't seem to be active (it's open from more than one year), maybe we can implement something on our own. Looking for some docs, I found a couple of examples, and it may be just a matter of properly setting up `logging.Formatter` https://stackoverflow.com/a/50146869/1303090

rokroskar:
Yes indeed the issue seems dead - but maybe this could be contributed to upstream. See the binderhub issue that is linked there.

lorenzo-cavazzi:
I created a separate issue for generating JSON log (#226) since this is a blocker for a couple of UI fixes

PR #243, opened 2020-02-07
fix: check lfs setting as a string

### LINKED ISSUE #227: LFS data is not fetched even if the option is selected
The server option `lfs_auto_fetch: true` is currently ignored when launching a new notebook server. See https://github.com/SwissDataScienceCenter/renku-ui/issues/736.

lorenzo-cavazzi:
The environment variable `LFS_AUTO_FETCH` is set to the string `"True"` when the lfs fetch checkbox is ticked: this is why the condition was always testing false when checking for `= 1`, therefore ignoring the setting.

I am not sure if this changed recently or if it simply didn't work in the past. Maybe we could fix this by making `LFS_AUTO_FETCH = true` (or `1`) instead of testing for `== "True"`

fix #227

***TESTING***
Preview available here: https://lorenzotest.dev.renku.ch/
* try to start notebooks on a project with lfs files, once with and once without selecting the lfs fetch checkbox
* verify with `git lfs ls-files` that lfs files are available

### DIFF
--- a/git-clone/entrypoint.sh
+++ b/git-clone/entrypoint.sh
@@ -13,7 +13,7 @@ fi
 
 rm -rf ${MOUNT_PATH}/*
 (rm -rf ${MOUNT_PATH}/.* || true)
-git config push.default simple
+git config --system push.default simple
 git lfs install $LFS_SKIP_SMUDGE --system
 git clone $REPOSITORY ${MOUNT_PATH}
 git lfs install $LFS_SKIP_SMUDGE --local
--- a/git-clone/entrypoint.sh
+++ b/git-clone/entrypoint.sh
@@ -5,7 +5,7 @@
 # the proper file permissions.
 set -x
 
-if [ "$LFS_AUTO_FETCH" == "True" ]; then
+if [ "$LFS_AUTO_FETCH" = 1 ]; then
   LFS_SKIP_SMUDGE="";
 else
   LFS_SKIP_SMUDGE="--skip-smudge";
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi:
@rokroskar I almost forgot that this morning we discussed a different solution that involves always using `--skip-smudge` and explicitly fetching when needed :man_facepalming: 
A code stub is the following.
```
git lfs install --skip-smudge --system
git clone $REPOSITORY ${MOUNT_PATH}
git lfs install --skip-smudge --local
if [ "$LFS_AUTO_FETCH" == "True" ]; then
  git lfs pull
fi
```
It seems to be better than the suggested fix and it is also more coherent with the new `renku clone` command that uses the `--skip-smudge` flag by default (if not explicitly told otherwise).
If it makes sense, I will submit another PR with this solution

rokroskar:
> The environment variable LFS_AUTO_FETCH is set to the string "True"

You mean in the UI? This must have changed recently I suppose, I'm pretty sure this option worked in the past. But anyway, setting an environment variable to `1` rather than `True` seems more standard to me. If `True` then I would at least make it lowercase.

lorenzo-cavazzi:
From the UI, the value seems to be a boolean and the value is `true`

Screenshot_20200210_083137

Anyway, I'll double-check what's going on to understand where that boolean becomes a capitalized  `"True"`

lorenzo-cavazzi:
Testing for `1` is definitely better than `"True"`, I addressed that.

Preview updated on https://lorenzotest.dev.renku.ch/

PR #253, opened 2020-03-03
feat(jupyterhub): update to version 1.1.0

### LINKED ISSUE #183: Upgrade JupyterHub to version 1.x

lorenzo-cavazzi:
This PR upgrade the JupyterHub version from `0.9.6` to `1.1.0` (last stable release).
It addresses a few stability/usability issues and it seems to enable the possibility to re-introduce the "force stop" button on environments (with some limitations -- killing pods on kubernetes still mess up the hub DB, but now it's only temporarily).

A test version is available here: https://lorenzotest.dev.renku.ch/

fix #183


jupyterhub/requirements.txt
```diff
@@ -16,7 +16,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Installing from Git until a new version of kubespawner (introducing 'delete_grace_period') is released.
--e git://github.com/jupyterhub/kubespawner.git@46a4b109c5e657a4c3d5bfa8ea4731ec6564ea13#egg=jupyterhub-kubespawner
-jupyterhub==0.9.6
+jupyterhub-kubespawner==0.11.1
```
rokroskar: 👍 thanks!

PR #261, opened 2020-03-12
feat: add requested resources info to servers endpoint

### LINKED ISSUE #223: feat: add resource information to pod annotations
It should be easy to add a couple of extra annotations to the pods about the environment resources/options selected during the start phase. These could be added to the `/servers` endpoint and be used by the UI to give extra information (e.g. number of GPU available)

lorenzo-cavazzi:
Add the requested resources info to the `GET /servers` API response. This is an example, the new part is in the attribute `resources`

```
{
  "servers": {
    "flights-tutorial-d991bad2": {
      "annotations": {
        [...]
      },
      "name": "flights-tutorial-d991bad2",
      "resources": {
        "cpu": "0.1",
        "memory": "1073741824"
      },
      "started": "2020-03-12T16:54:30.000000Z",
      "state": {
        "pod_name": "jupyter-lorenzo-2ecavazzi-2etech-flights-2dtutorial-2dd991bad2"
      },
      "status": {
        [...]
      },
      "url": "https://lorenzo.dev.renku.ch/jupyterhub/user/lorenzo.cavazzi.tech/flights-tutorial-d991bad2/"
    }
  }
}
```

UPDATE: It's possible to test the change in `lorenzotest` deplyoment: https://lorenzotest.dev.renku.ch/

fix #223

### DIFF
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -159,6 +159,23 @@ def _get_all_user_servers(user):
         status.update(conditions_summary)
         return status
 
+    def get_pod_resources(pod):
+        try:
+            resources = pod.spec.containers[0].resources.requests
+            # translate the cpu weird numeric string to a normal number
+            # ref: https://kubernetes.io/docs/concepts/configuration/
+            #   manage-compute-resources-container/#how-pods-with-resource-limits-are-run
+            if (
+                "cpu" in resources
+                and isinstance(resources["cpu"], str)
+                and str.endswith(resources["cpu"], "m")
+                and resources["cpu"][:-1].isdigit()
+            ):
+                resources["cpu"] = str(int(resources["cpu"][:-1]) / 1000)
+        except (AttributeError, IndexError):
+            resources = {}
+        return resources
+
     def get_server_url(pod):
         url = "/jupyterhub/user/{username}/{servername}/".format(
             username=pod.metadata.annotations["hub.jupyter.org/username"],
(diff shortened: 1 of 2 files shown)


rokroskar:
Thanks @lorenzo-cavazzi! I'm wondering if it would make sense to show both resource request and limit? otoh it might make the most sense to _enforce_ both the memory and cpu requests as limits. At the moment, only the memory is enforced. Then we only need to show a single number, like you have done here.

lorenzo-cavazzi:
We don't get any `cpu` value in the limits object if we don't enforce one. Currently, we only get the `memory` limit and it's always equal to the request counterpart.
On the other hand, do we *want* to enforce cpu limits? Is there any advantage? I see how allocating extra memory may be a problem when we need to reclaim that, but for the cpu I don't see the same problem

rokroskar:
> We don't get any cpu value in the limits object if we don't enforce one

yes, that was my point - maybe we should. What can happen is that one user can starve a node of CPU. CPU is a compressible resource, meaning the OS can decide to throttle it or not. If you don't enforce CPU limits, a single greedy process can take over all of the CPU on the node adversely affecting all others.

lorenzo-cavazzi:
I was going to look into enforcing the CPU limit since your argument is convincing. Should I open a different issue to keep track of that?

rokroskar:
You just need to add the CPU limit to the pod in the spawner in the same way it's done for the memory. But maybe it warrants a discussion with the others - although the decision is either: a) we hope that the current way doesn't make too many people unhappy or b) we enforce it and then document the limitations

PR #274, opened 2020-03-23
fix: use registry API to look up if image exists

### LINKED ISSUE #273: Interactive environments do not launch with the correct image
Because of pagination problems (see #272) the notebook service fails to find the correct image build job and therefore launches the notebook server with the wrong image. The solution should use the registry API instead (https://python-gitlab.readthedocs.io/en/stable/gl_objects/repository_tags.html).

rokroskar:
closes #273 

running on rok.dev.renku.ch

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -32,6 +32,7 @@ notebook = "*"
 pytest-mock = "*"
 jupyterhub-traefik-proxy = "==0.1.4"
 ptvsd = "~=4.2"
+importlib-metadata = "*"
 
 [requires]
 
--- a/renku_notebooks/util/gitlab_.py
+++ b/renku_notebooks/util/gitlab_.py
@@ -70,49 +70,37 @@ def get_notebook_image(user, namespace, project, commit_sha):
     """Check if the image built by GitLab CI is ready."""
     gl_project = get_project(user, namespace, project)
 
-    image = os.getenv("NOTEBOOKS_DEFAULT_IMAGE", "renku/singleuser:latest")
+    default_image = os.getenv("NOTEBOOKS_DEFAULT_IMAGE", "renku/singleuser:latest")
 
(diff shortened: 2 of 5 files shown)


renku_notebooks/util/gitlab_.py
```diff
@@ -70,41 +70,39 @@ def get_notebook_image(user, namespace, project, commit_sha):
...
     gl_project = get_project(user, namespace, project)
 
-    image = os.getenv("NOTEBOOKS_DEFAULT_IMAGE", "renku/singleuser:latest")
+    default_image = os.getenv("NOTEBOOKS_DEFAULT_IMAGE", "renku/singleuser:latest")
 
     commit_sha_7 = commit_sha[:7]
-
-    for pipeline in gl_project.pipelines.list():
-        if pipeline.attributes["sha"] == commit_sha:
-            status = _get_job_status(pipeline, "image_build")
```
mohammad-sdsc: The `_get_job_status` function can be removed.

rokroskar: Done!

rokroskar:
btw, this fix desperately needs a test that would verify that the correct image is being used - I will make a follow-up issue for it, it seems to me it's not straightforward, but I also don't have the test-making skills of @mohammad-sdsc :)

Pipfile
```diff
@@ -32,6 +32,7 @@ notebook = "*"
 pytest-mock = "*"
 jupyterhub-traefik-proxy = "==0.1.4"
 ptvsd = "~=4.2"
+importlib-metadata = "*"
```
lorenzo-cavazzi: where is this used?

rokroskar: it's not, but pytest fails without it

rokroskar:
🤞

PR #278, opened 2020-03-26
feat: contrain the container gids

rokroskar:
This PR constrains the gids used for the container processes. It's currently running on rok.dev.renku.ch. 

With this PR, the notebook container runs as gid 100 and sets the user to be a member of groups 100 and 1000.


rokroskar:
yes - this is fixed in the new images - the old images will have this annoying message for the time being. I'm not sure it's a huge problem... ?

rokroskar:
Thanks for testing @lorenzo-cavazzi!

PR #285, opened 2020-04-03
feat: allow unprivileged and anonymous users to launch servers

### LINKED ISSUE #171: make it possible to spawn notebook servers from public repository without logging in
I would like to be able to spawn a notebook server, as an anonymous user (not a project team member) of a public project. This is for tutorials and classes, we do not need (should not be able) to to push changes done in the notebook.

### LINKED ISSUE #275: Allow session launch for unprivileged (but logged in) users
The notebook launch sequence should be altered such that a logged in user without write access to a repo can still launch an interactive session. In this case, no oAuth token should be included in the repo URL.

ableuler:
- Enable the use of tmpauthenticator for Jupyterhub authentication.
  This enables notebook sessions for logged-out users. Closes #171

- Enable logged in users without write access to a project to start
  notebook servers from the project. Closes #275

### DIFF
--- a/helm-chart/make-values-tmp.py
+++ b/helm-chart/make-values-tmp.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+
+import argparse
+import secrets
+import sys
+
+import ruamel.yaml
+
+yaml = ruamel.yaml.YAML()
+yaml.preserve_quotes = True
+
+# Note: We hardcode "notebooks.fullname" / the service
+# name here. These secondary deployments MUST run in their
+# own dedicated namespace anyway.
+TEMPORARY_NOTEBOOKS_SERVICE_NAME = "notebooks-tmp"
+
+
+def main():
+    """Create a values file for a secondary notebooks service helm deployment
+    (inluding a secondary Jupyterhub) which handles temporary sessions for
+    logged-out users."""
+
+    argparser = argparse.ArgumentParser(description=main.__doc__)
+
+    argparser.add_argument("--values-file", help="Input values file", required=True)
+    argparser.add_argument(
+        "--renku-namespace", help="Namespace for the Renku deployment", required=True
+    )
+    argparser.add_argument(
+        "--output", "-o", help="Output file", default="notebooks-tmp-values.yaml"
+    )
+    args = argparser.parse_args()
+
+    with open(args.values_file) as f:
+        values = yaml.load(f.read())
+
+    # Copy over all the necessary stuff from the original values file
+    new_values = values["notebooks"]
+
+    # We don't know the release name here, so we override the name of the service
+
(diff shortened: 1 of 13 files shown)


jupyterhub/spawners.py
```diff
@@ -87,51 +101,49 @@ def start(self, *args, **kwargs):
...
-            )
-            self.log.debug(
-                "access level for user {username} in "
-                "{namespace}/{project} = {access_level}".format(
-                    username=self.user.name,
-                    namespace=namespace,
-                    project=project,
-                    access_level=access_level,
-                )
-            )
```
lorenzo-cavazzi, line 120: should we keep logging similar info? It may be useful to have something like that when debugging

helm-chart/renku-notebooks/values.yaml
```diff
@@ -98,9 +98,8 @@ jupyterhub:
       tag: 'latest'
     allowNamedServers: true
     services:
-      gateway:
-        oauth_no_confirm: true
       notebooks:
+        oauth_no_confirm: true
```
rokroskar, line 102: was this just in the wrong place before?

helm-chart/make-values-tmp.py
```diff
@@ -0,0 +1,100 @@
...
+            "valueFrom": {
+                "secretKeyRef": {
+                    "name": "renku-jupyterhub-tmp-postgres",
+                    "key": "jupyterhub-tmp-postgres-password",
+                }
+            },
+        }
+    )
+    hub_section["extraEnv"].append({"name": "JUPYTERHUB_AUTHENTICATOR", "value": "tmp"})
+    hub_section["services"]["notebooks"]["url"] = "http://{}-renku-notebooks".format(
```
lorenzo-cavazzi, line 70: this produces an url like this: `http://<namespace>-tmp-renku-notebooks`
but the correct links seems to be `http://<namespace>-renku-notebooks-tmp`

jupyterhub/spawners.py
ableuler: Good point, added it back.

helm-chart/renku-notebooks/values.yaml
ableuler: No. For the gateway I have moved this information to the renku chart because there is no gateway service registered with the Jupyterhub when running as a secondary deployment for anonymous sessions (see our long discussion on the Renku chart ;). 
Why the missing `oauth_no_confirm: true` on the notebooks service didn't result in any confirmation screen showing up in our "regular" flow I don't know, but when using the tmpauthenticator it is needed.

helm-chart/make-values-tmp.py
ableuler: Thanks for spotting this. It's a shortcoming of helm that there is no way to refer to the release name in a values.yaml file, so this was a hopeful guess at best. I've replaced it by setting the name of that service to `notebooks-tmp` always.

Issue #280, opened 2020-03-27
Deactivate autosave hooks for unprivileged sessions.


ableuler:
In cases where the access level of a user does not allow opening a new branch and pushing to it, the autosave mechanisms should be disabled.

PR #289, opened 2020-04-15
feat: autosave/restore only when session is authenticated

mohammad-sdsc:
Fixes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/280

## Note
My understanding was that in anonymous sessions `JUPYTERHUB_AUTHENTICATOR` is set to a value other than `gitlab`. I used this to distinguish anonymous sessions from authenticated ones.

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -144,16 +144,20 @@ class SpawnerMixin:
                 )
             )
 
-            if (
-                len(access_levels) > 0
-                and max(access_levels) >= gitlab.MAINTAINER_ACCESS
-            ):
+            access_level = gitlab.GUEST_ACCESS
+            if len(access_levels) > 0:
+                access_level = max(access_levels)
+
+            if access_level >= gitlab.MAINTAINER_ACCESS:
 
                 environment = {
                     variable.key: variable.value
                     for variable in gl_project.variables.list()
                 }
                 self.environment.update(environment)
+
+            if access_level >= gitlab.DEVELOPER_ACCESS:
+                self.environment["GITLAB_AUTOSAVE"] = "1"
 
         result = yield super().start(*args, **kwargs)
         return result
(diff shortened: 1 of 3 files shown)


mohammad-sdsc:
> There are two cases where the autosave should be disabled:
> 
> * User is logged in, but does not have developer access to the project.
> * User is not logged in.
> 
> Currently you are covering the second case, except that the `JUPYTERHUB_AUTHENTICATOR` env variable is not defined in the init container nor the notebook container. So what I suggest is to define a `GITLAB_AUTOSAVE` environment variable in the spawner for the init- and notebook container. Then you can check for the value of this variable in the same place where you are doing it already.

It checked for the case where `JUPYTERHUB_AUTHENTICATOR` is not defined but not for the non-developer logged-in users. I've updated the PR.

jupyterhub/spawners.py
```diff
@@ -144,17 +144,21 @@ def start(self, *args, **kwargs):
...
+            if access_level >= gitlab.MAINTAINER_ACCESS:
 
                 environment = {
                     variable.key: variable.value
                     for variable in gl_project.variables.list()
                 }
                 self.environment.update(environment)
 
+            if access_level >= gitlab.DEVELOPER_ACCESS:
+                self.environment["GITLAB_AUTOSAVE"] = "1"
```
ableuler, line 160: This will only add it to the environment variables of the notebook container. You have to explicitly add it to the env variables of the init container too (see line 218).

mohammad-sdsc: Thanks @ableuler for noticing this! It's been a while that I've touched this code base!

mohammad-sdsc:
I've tested this on my dev env with and without developer access and it worked as expected.

mohammad-sdsc:
One thing that I've noticed is that UI shows the "Autosaved data recovered ..." prompt even if the user is anonymous or does not have developer access (i.e. nothing with be restored/autosaved). We should disable that in UI.

jupyterhub/spawners.py
ableuler: No problem. I spent many of my quarantine days staring at this, so I knew what to look out for... 😂

ableuler:
> One thing that I've noticed is that UI shows the "Autosaved data recovered ..." prompt even if the user is anonymous or does not have developer access (i.e. nothing with be restored/autosaved). We should disable that in UI.

Good catch. Can you create a UI issue for this please?

PR #291, opened 2020-04-17
Fix for gitlab URL

pameladelgado:
This is needed for some deployments that dont have `gitlab.url` configured


ableuler:
> This is needed for some deployments that dont have `gitlab.url` configured

Yes, but let's fix it here nevertheless. Thanks!

PR #297, opened 2020-04-24
Prep 0.7.2 release

ableuler:


rokroskar:
This should have been 0.8.0 I guess...?

ableuler:
> This should have been 0.8.0 I guess...?

On the notebooks-service the introduced changes _should_ actually be backwards-compatible (I admit that we haven't tested this aspect very thoroughly). That's why I went for `0.7.2`.

rokroskar:
sure but it's (much) more than a bugfix.

ableuler:
yes, but we have been doing this all the time, no? I thought before 1.0.0 we were following [this logic](https://github.com/dbrock/semver-howto#alpha-and-beta-0xy), but we can also agree to increase the second digit for new features.

PR #327, opened 2020-06-08
use user credentials for pulling images

### LINKED ISSUE #105: Use user credentials for pulling private images
Currently the gitlab sudo token is still used for pulling private images. Instead we should use the users credentials. Once this is done, the gitlab sudo token is not used anymore and can thus be removed from the renku chart.

rokroskar:
This PR does away with the need for admin access to the gitlab instance that backs the renku deployment by using the oauth2 token of the user launching the environment to authenticate with the image registry. 

closes #105


renku_notebooks/util/kubernetes_.py
```diff
@@ -212,3 +213,63 @@ def read_namespaced_pod_log(pod_name, max_log_lines=0):
...
+            "annotations": {"renku.io/username": user.get("name")},
+            "labels": {
+                "component": "singleuser-server",
+                "renku.io/username": user.get("name"),
+            },
+        },
+        type="kubernetes.io/dockerconfigjson",
+    )
+
+    existing = _check_existing_secret(user, kubernetes_namespace)
```
ableuler, line 254: You could just use `v1.read_namespaced_secret(namespace, name)` which corresponds to `GET /api/v1/namespaces/{namespace}/secrets/{name}`.

rokroskar: haha yep I could just do that - I massively over-complicated this. Not sure why.

PR #346, opened 2020-07-09
chore: obscure oauth token

### LINKED ISSUE #313: obscure the oauth2 token
At the moment we are injecting the oauth2 token into the repository remote - this is not ideal because when users get an error and the remote is displayed in the error string they copy paste the whole thing essentially broadcasting their password to the world. There is no way for us to obscure this in a predictable way, but we could put the credentials into the git config instead of the remote. 

Relevant git documentation: https://git-scm.com/docs/gitcredentials

rokroskar:
This PR restructures the git-clone entrypoint to inject the user's credentials into a credential-helper file so as to hide the token from the remote URL. 

This configuration is running on https://rok.dev.renku.ch

To test, launch an environment in the test deployment above and try to push a chance to one of your projects. See that `git remote -v` shows no oauth token. 

closes #313

### DIFF
--- a/git-clone/entrypoint.sh
+++ b/git-clone/entrypoint.sh
@@ -11,12 +11,29 @@ else
   LFS_SKIP_SMUDGE="--skip-smudge";
 fi
 
+# clear path
 rm -rf ${MOUNT_PATH}/*
 (rm -rf ${MOUNT_PATH}/.* || true)
+
+# set up git defaults
 git config --system push.default simple
-git lfs install $LFS_SKIP_SMUDGE --system
-git clone $REPOSITORY ${MOUNT_PATH}
+
+# extract the GitLab host and path
+pat='^(http[s]?:\/\/)([^:\/\s]+)\/?([a-zA-Z0-9_]+)?$'
+[[ $GITLAB_URL =~ $pat ]]
+GITLAB_HOST="${BASH_REMATCH[2]}"
+GITLAB_PATH="${BASH_REMATCH[3]}"
+
+# set up the repo
+mkdir -p $MOUNT_PATH
+cd $MOUNT_PATH
+git init
 git lfs install $LFS_SKIP_SMUDGE --local
+git config credential.helper "store --file=.git/credentials"
+echo "https://oauth2:${GITLAB_OAUTH_TOKEN}@${GITLAB_HOST}" > .git/credentials
+git remote add origin $REPOSITORY
+git fetch origin
+git checkout master
 
 if [ "${GITLAB_AUTOSAVE}" == "1" ] ; then
 
--- a/git-clone/entrypoint.sh
+++ b/git-clone/entrypoint.sh
@@ -18,7 +18,7 @@ rm -rf ${MOUNT_PATH}/*
 git config --system push.default simple
 
 # extract the GitLab host and path
-pat='^(http[s]?:\/\/)([^\/]+)\/?([a-zA-Z0-9_\/\-]+)?$'
+pat='^(http[s]?:\/\/)([^\/]+)\/?([a-zA-Z0-9_\/\-]+?)$'
 [[ $GITLAB_URL =~ $pat ]]
 GITLAB_HOST="${BASH_REMATCH[2]}"
 GITLAB_PATH="${BASH_REMATCH[3]}"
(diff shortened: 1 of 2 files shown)


git-clone/entrypoint.sh
```diff
@@ -11,12 +11,29 @@ else
...
 rm -rf ${MOUNT_PATH}/*
 (rm -rf ${MOUNT_PATH}/.* || true)
+
+# set up git defaults
 git config --system push.default simple
-git lfs install $LFS_SKIP_SMUDGE --system
-git clone $REPOSITORY ${MOUNT_PATH}
+
+# extract the GitLab host and path
+pat='^(http[s]?:\/\/)([^:\/\s]+)\/?([a-zA-Z0-9_]+)?$'
```
ableuler, line 22:
```suggestion
pat='^(http[s]?:\/\/)([^\/]+)\/?([a-zA-Z0-9_\/\-]+)?$'
```
This would also correctly match the following:
`https://some.host:port/git-lab/path`

ableuler:
Great to have the oauth token out of the repo URL! 👍

rokroskar:
> P.S: remember to update the UI version to the latest.

not sure what you mean? It was working fine for me? I was using the master from renku from 1-2 days ago.

lorenzo-cavazzi:
@rokroskar I couldn't start notebooks from the latest commits of non-master branches (the commits weren't updating on branch selection, UI bug)

rokroskar:
Aha ok, maybe I'm a few commits behind then. In any case that sounds to be out of the scope of this PR - or were you just trying to make sure that the branches are still handled correctly with this change?

git-clone/entrypoint.sh
rokroskar: do we want this to be more generic?

lorenzo-cavazzi:
Yes, I was checking that the branch was correctly checked out because I saw a `git checkout master`

rokroskar:
Actually that's a good point - I added that when I was testing and wanted to make it more generic but now I can't find where the requested branch was checked out before...

git-clone/entrypoint.sh
ableuler: I would say so. Forbidding hyphens or slashes in the path seems like an unnecessary restriction to me, and allowing different ports might be helpful in some dev settings.

rokroskar: yeah but then do you want a separate capture group for the port?

rokroskar:
I just pushed a version with the recommended changes and UI 0.10.3. Could you please check that it's working fine? If so and you approve the PR, please go ahead and merge it.

git-clone/entrypoint.sh
```diff
@@ -3,20 +3,35 @@
...
 rm -rf ${MOUNT_PATH}/*
 (rm -rf ${MOUNT_PATH}/.* || true)
+
+# set up git defaults
 git config --system push.default simple
-git lfs install $LFS_SKIP_SMUDGE --system
-git clone $REPOSITORY ${MOUNT_PATH}
+
+# extract the GitLab host and path
+pat='^(http[s]?:\/\/)([^\/]+)\/?([a-zA-Z0-9_\/\-]+)?$'
```
ableuler, line 21:
```suggestion
pat='^(http[s]?:\/\/)([^\/]+)\/?([a-zA-Z0-9_\/\-]+?)$'
```
The bash version in alpine linux seems to give preference to greedy matches closer to the end of the string, while the version in ubuntu or on mac is "greedier" at the beginning. By making the path regex lazy while the hostname regex remains greedy, the problem is fixed.

rokroskar:
@lorenzo-cavazzi @ableuler I've updated the deployment and tested that I can push from an interactive session.

rokroskar:
Still needs @lorenzo-cavazzi approval - note that the deployed version on rok.dev.renku.ch is no longer this PR... but I can bring it back if needed.

lorenzo-cavazzi:
Not sure if I did something wrong while creating the image locally, but when I try to upgrade the deployment, the `hub` component disappears. Would you mind trying to update https://rok.dev.renku.ch just to be sure everythig is working?

rokroskar:
How are you doing the update? If you are using the notebooks chart from a local directory make sure you run `helm dep update` in the `renku-notebooks` repo _before_ running `helm dep update` in the main renku repo.

rokroskar:
Thanks @lorenzo-cavazzi!

PR #355, opened 2020-07-27
chore: update jh chart version

rokroskar:
Updates the JH chart dependency.


rokroskar:
the JH version remains the same, but no, I haven't removed it from the `requirements.txt`

PR #377, opened 2020-08-25
fix: update lfs version in git-clone to correctly handle stored credentials

### LINKED ISSUE #375: Fetching lfs files on environment launch fails for private projects
This is the error which is logged in the init container:

```
Downloading data/<redacted>.csv (285 MB)
Error downloading object: data/<redacted>.csv (<short-sha>): Smudge error: Error downloading data/<redacted>.csv (<full-sha>): LFS: Git credentials for <object-url>?temp_url_sig=<redacted>&temp_url_expires=1598283623 not found.
```

ableuler:
This fixes a bug which occurred when pulling lfs data from a private repo in the init-container by upgrading the git alpine version and thereby the git lfs version to `2.11.0`. When GitLab is configured do redirect to the object store for getting the data (rather than proxying to it), the previous version of git lfs expected to find credentials for the object store itself in the local credential store, which are of course missing. The new version of git lfs does not suffer from that problem.

closes #375


rokroskar:
Is there a specific change in LFS that you found that addressed this? Was this a bug in LFS?

ableuler:
> Is there a specific change in LFS that you found that addressed this? Was this a bug in LFS?

I had a quick look and didn't find one. There are a few fixes related to the handling of credentials, but none that is obviously linked to our issue.

rokroskar:
ok - I was just wondering if we can expect it to happen again

ableuler:
> ok - I was just wondering if we can expect it to happen again

I don't think so. If you ask me it clearly _was_ a bug in lfs. The LFS server sends a redirect to a temporary object store URL where the file can be downloaded (which does not require any authentication as far as I understand). lfs expecting to find locally stored credentials for this object store URL (and failing if it doesn't) is clearly a bug. Fixing it might have been as easy as trying without credentials if it can't find any.
It was probably fixed as part of the 2.6 release, but I just didn't find a specific issue.
```
This release adds better support for redirecting network calls from a Git LFS API server to one that requires a different authentication mode, builds Git LFS on Go 1.11, and numerous other bug fixes and modifications.
```

rokroskar:
ok - I just meant that if it wasn't specifically fixed then there might not be regressions tests for it in which case we could expect this to break in the future. In which case it would be  worth opening an issue with LFS.

rokroskar (approved):

ableuler:
> ok - I just meant that if it wasn't specifically fixed then there might not be regressions tests for it in which case we could expect this to break in the future. In which case it would be worth opening an issue with LFS.

ok, I see what you mean. Yes, I guess that probability might still be there. I created an [issue on lfs](https://github.com/git-lfs/git-lfs/issues/4223) asking about it.

rokroskar:
oops we forgot to bump the LFS version used in the images...

ableuler:
> oops we forgot to bump the LFS version used in the images...

Is it causing any problems? The one use in the images is newer than one we used to have in the init container and it seemed to correctly handle the credentials. Therefore I saw no immediate need to upgrade the version in the images.

rokroskar:
I thought it was a source of problems for some users yesterday but it turned out to be something else. But isn't it going to cause failures for users who might have slightly older images?

ableuler:
Could be - I didn't check if older base images use a problematic version of lfs.

rokroskar:
Which version is needed?

ableuler:
most likely `>=2.6`. I remember `2.8` to work and I think it was `2.5` that did not work.

rokroskar:
Ok great - we've had 2.8 for some time I think

PR #384, opened 2020-09-01
fix(charts): correct proxy-public port in ingress

ableuler:
Same as https://github.com/SwissDataScienceCenter/renku/pull/1446.


pameladelgado (approved):
:+1:

PR #397, opened 2020-09-11
chore(charts): add reference for gw client app to values

ableuler:


helm-chart/renku-notebooks/values.yaml
```diff
@@ -101,6 +101,12 @@ jupyterhub:
         oauth_client_id: service-notebooks
         ## Generate the notebooks service api token with `openssl rand -hex 32`
         # api_token: notebookstoken
+      gateway:
+        admin: true
+        oauth_client_id: gateway
+        ## MUST BE IDENTICAL TO gateway.jupyterhub.clientSecret
+        # apiToken: Generat with `openssl rand -hex 32`
```
rokroskar, line 108:
```suggestion
        # apiToken: Generate with `openssl rand -hex 32`
```

rokroskar (approved):

PR #399, opened 2020-09-12
chore(charts): several improvements

ableuler:
Some small improvements to the charts (see commit messages).

### DIFF
--- a/helm-chart/minikube-values.yaml
+++ b/helm-chart/minikube-values.yaml
@@ -54,6 +54,3 @@ jupyterhub:
     gitlab:
       clientId: 75be9e2e28203c3c44272c72142a839956cfc07e1a0ab93a5d7c9ac6c7802b58
       clientSecret: 4bd8a67690d4a102436d15fa82b793d73046d350372ee6bd0302617298d08092
-
-securityContext:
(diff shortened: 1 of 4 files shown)


rokroskar:
It's good to simplify this, but I'm not sure I understand - doesn't telepresence always create its own deployment anyway?

ableuler:
> It's good to simplify this, but I'm not sure I understand - doesn't telepresence always create its own deployment anyway?

No, not anymore (https://github.com/telepresenceio/telepresence/pull/1404). My interpretation was that when using it's own deployment, telepresence would not be bound to the securityContext set on the original deplouyment. But to be honest, I don't claim to fully understand this. I simply tried with the latest telepresence `0.108` against a deployment which had the securityContext enabled and realized that setting `TELEPRESENCE_USE_DEPLOYMENT=1` made it work. But since this has been the default behaviour until recently, I'm not sure why you had to introduce this distinction on the chart in the first place.

rokroskar:
I can't remember the reason any more, but it didn't use to work with the security context in place. It may have been something to do with access to the service account credentials.

rokroskar (approved):

PR #417, opened 2020-10-02
feat: enable pod autoscaling

rokroskar:
Provides an optional HPA template to enable autoscaling, configurable through the values file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,6 @@ repos:
     rev: stable
     hooks:
     - id: black
-      language_version: python3.7
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.2.3
     hooks:
--- a/helm-chart/renku-notebooks/templates/hpa.yaml
+++ b/helm-chart/renku-notebooks/templates/hpa.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "notebooks.fullname" . }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "notebooks.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
+  targetCPUUtilizationPercentage: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+{{- end }}
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -24,9 +24,9 @@ replicaCount: 1
 ## Configure autoscaling
 autoscaling:
   enabled: false
-#   minReplicas: 1
-#   maxReplicas: 5
-#   targetCPUUtilizationPercentage: 50
+  minReplicas: 1
+  maxReplicas: 5
+  targetCPUUtilizationPercentage: 50
 
 image:
   repository: renku/renku-notebooks
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -21,6 +21,13 @@ gitlab:
 
 replicaCount: 1
 
+## Configure autoscaling
+autoscaling:
+  enabled: false
+#   minReplicas: 1
+#   maxReplicas: 5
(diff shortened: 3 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -3,7 +3,6 @@ repos:
     rev: stable
     hooks:
     - id: black
-      language_version: python3.7
```
ableuler, line 6: Is this change on purpose? Just asking as it seems unrelated to the rest of this PR.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -21,6 +21,13 @@ gitlab:
 
 replicaCount: 1
 
+## Configure autoscaling
+autoscaling:
+  enabled: false
```
ableuler, line 26: I'm wondering if we should fully adopt the bitnami convention and have two values files in each chart. A `values.yaml` and a `production-values.yaml` which would have autoscaling on per default and a minimum of two replicas. Or maybe it's enough to do this on the RenkuLab chart level?

helm-chart/renku-notebooks/templates/hpa.yaml
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "notebooks.fullname" . }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "notebooks.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
```
ableuler, line 11: Do you know if the HPA guarantees that the replicas are spread out over at least two nodes?

ableuler (commented):
Great and long-needed addition. The main further improvement I see would be to use `autoscaling/v2beta2` instead of `autoscaling/v1` and add autoscaling also based on memory usage. But if you don't want to tackle this right now I'm happy to merge the PR "as-is" as this is already a great improvement over the status quo.

.pre-commit-config.yaml
rokroskar: the pre-commit hook was failing because I'm using python 3.8

helm-chart/renku-notebooks/values.yaml
rokroskar: yes, I think it's enough to do it there

helm-chart/renku-notebooks/templates/hpa.yaml
rokroskar: pretty sure it just hands the request to the scheduler - you'd need to have anti-affinity set separately. This isn't about HA I guess

rokroskar:
Why would you scale the notebooks service based on memory usage? I think the memory usage should be pretty constant.

helm-chart/renku-notebooks/templates/hpa.yaml
pameladelgado: I think it does

pameladelgado (dismissed):
I tested this and it works wonderfully.

ableuler:
I agree that I'd expect the CPU usage to scale more linearly with the number of requests per time, but I think that memory based autoscaling can make our services a bit more robust against things like memory leaks, requests piling up because of slow upstream services etc.

rokroskar:
Thanks for testing @pameladelgado - I pushed a change to make some sensible defaults so all that is needed is to set the `autoscaling.enabled=true` instead of having to set the whole block.

rokroskar:
> I agree that I'd expect the CPU usage to scale more linearly with the number of requests per time, but I think that memory based autoscaling can make our services a bit more robust against things like memory leaks, requests piling up because of slow upstream services etc.

Sure, good point - I'd say lets see if this is a possibility/issue and if so add a second metric.

pameladelgado:
> Thanks for testing @pameladelgado - I pushed a change to make some sensible defaults so all that is needed is to set the `autoscaling.enabled=true` instead of having to set the whole block.

Scaling down does take a while, I guess this is normal.
Regarding memory I was going to suggest the same as @ableuler but I also thought that for notebooks memory might not be a critical resource.

pameladelgado (approved):

helm-chart/renku-notebooks/templates/hpa.yaml
ableuler: You're right @pameladelgado. The scheduler in its default settings will try to spread out pods, but it won't guarantee anything (it's just entering the big priority scoring function): 
```
SelectorSpreadPriority: Spreads Pods across hosts, considering Pods that belong to the same Service, StatefulSet or ReplicaSet.
```

PR #419, opened 2020-10-04
fix(app): escape username in registry secret name

### LINKED ISSUE #393: underscores in username result in illegal DNS name for user registry secrets
See the sentry issue:  [RENKU-NOTEBOOKS-D](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/233/?referrer=github_integration)

Related: https://github.com/SwissDataScienceCenter/renku/issues/1128

```
ApiException: (422)
Reason: Unprocessable Entity
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Fri, 04 Sep 2020 15:22:09 GMT', 'Content-Length': '886'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Secret \"maja_do-registry\" is invalid: metadata.name: Invalid value: \"maja_do-registry\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e...
(11 additional frame(s) were not displayed)
...
  File "kubernetes/client/api_client.py", line 345, in call_api
    _preload_content, _request_timeout)
  File "kubernetes/client/api_client.py", line 176, in __call_api
    _request_timeout=_request_timeout)
  File "kubernetes/client/api_client.py", line 388, in request
    body=body)
  File "kubernetes/client/rest.py", line 278, in POST
    body=body)
  File "kubernetes/client/rest.py", line 231, in request
    raise ApiException(http_resp=r)
```

ableuler:
This PR fixes the problem that users with an underscore in their username have seen when launching notebooks from private repos.
Closes #393.

### DIFF
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -18,9 +18,10 @@
 """Notebooks service API."""
 import json
 import os
-from kubernetes.client.rest import ApiException
 
+import escapism
 from flask import Blueprint, abort, current_app, jsonify, request, make_response
+from kubernetes.client.rest import ApiException
 
 from .. import config
 from ..util.gitlab_ import (
@@ -146,7 +147,8 @@ def launch_notebook(user):
 
     # only create a pull secret if the project has limited visibility and a token is available
     if config.GITLAB_AUTH and gl_project.visibility in {"private", "internal"}:
-        secret_name = f"{user.get('name')}-registry"
+        safe_username = escapism.escape(user.get("name"), escape_char="-").lower()
+        secret_name = f"{safe_username}-registry"
         create_or_replace_registry_secret(user, namespace, secret_name)
         payload["image_pull_secrets"] = [secret_name]
 


ableuler:
An image which includes the fix is available under `ableuler/renku-notebooks:0.8.1-e20b7a0` for easier testing of this PR.

renku_notebooks/api/notebooks.py
```diff
@@ -146,7 +147,8 @@ def launch_notebook(user):
 
     # only create a pull secret if the project has limited visibility and a token is available
     if config.GITLAB_AUTH and gl_project.visibility in {"private", "internal"}:
-        secret_name = f"{user.get('name')}-registry"
+        safe_username = escapism.escape(user.get("name"), escape_char="-").lower()
+        secret_name = f"{safe_username}-registry"
```
rokroskar, line 151: Should this be handled inside `create_or_replace_registry_secret`? Or in a separate function? We'll need it again when removing the secret...

rokroskar (approved):
Just a tiny refactoring suggestion, otherwise 👍

renku_notebooks/api/notebooks.py
ableuler: It agree that once we tackle the removal, this logic will very likely end up in a separate helper function. But since it's not yet entirely clear to me how we will implement the removal (I think it would be nice to couple the lifetime of the secret as directly to the pod lifetime) I'd leave this for later.

PR #429, opened 2020-10-06
Disable userScheduler and userPlaceholder

pameladelgado:
To still fix/verify: disable image puller


rokroskar:
Could you also disable the continuous image puller?

ableuler (dismissed):
Thanks!

rokroskar (approved):

PR #431, opened 2020-10-09
Remove service links from user pods

### LINKED ISSUE #155: remove service links from notebook pod environment
see `enableServiceLinks` option in [podSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#pod-v1-core)

The usage of environment variables for accessing services is deprecated in k8s anyway and this exposes lots of information about services to users. This change will mean that the notebooks service requires k8s > 1.13

olevski:
Closes #155 

Number of env variables on renkulab.io freshly made project:
```
 work ❯ test1 ▶ master ▶ $ ▶ env | wc -l
258
```

After setting enableServiceLinks to false:
```
 work ❯ test10 ▶ master ▶ $ ▶ env | wc -l
66
```

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -299,7 +299,7 @@ class RenkuKubeSpawner(SpawnerMixin, KubeSpawner):
         # set the image pull policy
         self.image_pull_policy = "Always"
 
-        # Disable service links to prevent a lot of environment variables appearing in user environment
+        # Prevent kubernetes service links from appearing in user environment
         self.extra_pod_config = { "enableServiceLinks": False }
 
         pod = yield super().get_pod_manifest()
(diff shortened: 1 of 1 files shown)


ableuler (dismissed):
👍

jupyterhub/spawners.py
```diff
@@ -299,6 +299,9 @@ def get_pod_manifest(self):
         # set the image pull policy
         self.image_pull_policy = "Always"
 
+        # Disable service links to prevent a lot of environment variables appearing in user environment
```
rokroskar, line 302: Maybe a bit clearer to our future selves about the reason for this

```suggestion
        # Prevent kubernetes service links from appearing in user environment
```

ableuler (approved):

PR #458, opened 2020-11-16
Replace deprecated set-env commands in the GitHub actions

### LINKED ISSUE #439: chore: update deprecated github actions
Handling of environment variables in GitHub actions has changed due to a security vulnerability - see [here](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

We should update the CI in all repos to use the new syntax described [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files)

lorenzo-cavazzi:
fix #439

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,8 @@ jobs:
         fetch-depth: 0
     - name: set up environment variables
       run: |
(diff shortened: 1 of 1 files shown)


rokroskar (approved):
Thanks @lorenzo-cavazzi!!

PR #461, opened 2020-11-17
Fix cull secrets bug

### LINKED ISSUE #459: Fix ambiguity in registry-secret to pod matching
Currently, there are corner cases where two different user pods can refer to the same registry secret. This does result in a problem when cleaning up the secrets:

```
INFO:root:Checking for user registry secrets whose names match the regex: .+-registry-[a-z0-9-]{36}$
File "/cull_secrets/clean_user_registry_secrets.py", line 155, in <module>
main()
File "/cull_secrets/clean_user_registry_secrets.py", line 151, in main
remove_user_registry_secret(args.namespace, k8s_client, args.age_hours_minimum)
File "/cull_secrets/clean_user_registry_secrets.py", line 84, in remove_user_registry_secret
podname = find_pod_by_secret(secret, k8s_client)
File "/cull_secrets/clean_user_registry_secrets.py", line 47, in find_pod_by_secret
"There should at most one pod that matches a secret, "
Exception: There should at most one pod that matches a secret, found 2 that match the secret
```

olevski:
Fixes #459 

So the initial idea for this fix was to just add the GIT_URL (which is already a user server pod annotation) as a label for pod the user pods and the image pull secrets. However, the problem is that `/` are not allowed in k8s label values (but they are allowed in annotations). 

To bypass this I essentially decided to add all the parts of the repository url as separate labels. So now we have the following that identifies a unique server pod or image pull secret
- host
- namespace
- username
- commit sha
- project name

We have to use both the namespace and username because in the case where a user runs a session on a shared repo (that does not belong to them) but then also clones the repo and then runs another session on the same commit the only difference between the user session pods is in the namespace. I just tested this case which is what was causing the bug and the secret cleanup works. If you want to test let me know and I can give you access to one of my dummy test projects so that you can run a pod straight from it but then also fork it and run the same pod again.


jupyterhub/spawners.py
```diff
@@ -277,6 +277,9 @@ def get_pod_manifest(self):
             + "/"
             + options.get("project")
         )
+        git_host = urlparse(
+            os.environ.get("GITLAB_URL", "http://gitlab.renku.build"),
```
ableuler, line 281: I know we still have this elsewhere, but `http://gitlab.renku.build` as a default for development is deprecated. The code should fail if no `GITLAB_URL` is provided as environment variable.

cull_secrets/clean_user_registry_secrets.py
```diff
@@ -55,7 +61,13 @@ def find_pod_by_secret(secret, k8s_client):
 def remove_user_registry_secret(namespace, k8s_client, max_secret_age_hrs=0.25):
     """Used in a cronjob to periodically remove old user registry secrets"""
     secret_name_regex = ".+-registry-[a-z0-9-]{36}$"
-    label_keys = ["renku.io/commit-sha", "renku.io/projectName", "renku.io/username"]
+    label_keys = [
```
ableuler, line 64: You could define the `label_keys` at the module level and use it both locations. One day someone inattentive (like myself) will modify this list in only one of the two locations and produce a bug.

jupyterhub/spawners.py
```diff
@@ -277,6 +277,9 @@ def get_pod_manifest(self):
...
+        ).netloc
```
ableuler, line 282: Please add the `.path` here for the (admittedly unlikely) case that two gitlab instances are running under the same hostname, like `org.host/team-a/gitlab` and `org.host/team-b/gitlab`./

renku_notebooks/api/notebooks.py
```diff
@@ -148,10 +149,11 @@ def launch_notebook(user):
 
     # only create a pull secret if the project has limited visibility and a token is available
     if config.GITLAB_AUTH and gl_project.visibility in {"private", "internal"}:
+        git_host = urlparse(config.GITLAB_URL).netloc
```
ableuler, line 152: See comment above about `git_host`.

ableuler (changes requested):
Thanks for the fix!

jupyterhub/spawners.py
olevski: @ableuler I would have to add the first portion of the path as a separate label. The reason for this is that k8s label values cannot contain `/`. Is this what you are referring? I.e. now there will be an extra label just called `git-path` and it would also sometimes be blank because sometimes gitlab is not deployed on `hostname.com/gitlab` but rather at `gitlab.something.com`.

ableuler: > The reason for this is that k8s label values cannot contain `/`

Ah, I did not think about this. 

> Is this what you are referring?

No, I was just thinking that since you add all the components of the repo URL as labels (which I like), these labels should also allow to uniquely identify a repo. With the path of the git host missing, this is not strictly the case. Admittedly, this would only matter when one RenkuLab instance can work with multiple GitLab instances AND two of those GitLab instances are running under different paths but share the same domain. Maybe just add the path and URL-encode it?

jupyterhub/spawners.py
olevski: @ableuler I removed all of the instances of this wherever it appears in the whole repo. There were quite a few. I can roll them back without a problem if you did not want me to get them all.

jupyterhub/spawners.py
```diff
@@ -70,7 +70,7 @@ def get_env(self):
             "CI_NAMESPACE": self.user_options.get("namespace", ""),
             "CI_PROJECT": self.user_options.get("project", ""),
             "CI_COMMIT_SHA": self.user_options.get("commit_sha", ""),
-            "GITLAB_URL": os.environ.get("GITLAB_URL", "http://gitlab.renku.build"),
+            "GITLAB_URL": os.environ["GITLAB_URL"],
```
rokroskar, line 73: please use `os.environ.get` and handle the case of `None` - otherwise if it's not set you get an unhandled `KeyError` here.

rokroskar (changes requested):

jupyterhub/spawners.py
olevski: Let me know if you agree with how I fixed this. I only took the first part of the `path`. So the following happens:
- if `GIT_URL = 'https://gitlab.com'`, then path is `""`
- if `GIT_URL = 'https://hostname.com/gitlab'`, then path is `"gitlab"`
- if `GIT_URL = 'https://hostname.com/team1/gitlab'`, then path is `"team1"`

jupyterhub/spawners.py
olevski: @rokroskar see the first comment in this PR. I thought we wanted things to crash if we do not specify the GIT_URL as an environment variable.

rokroskar: Ah I missed that. Crashing with a KeyError is not the best way to handle a missing required config option imho - the app should fail already when `helm` tries to render the chart, but if deployed outside that and required config options are missing it should give a meaningful error and exit (in the cases where we can't provide a reasonable default).

ableuler: > Ah I missed that. Crashing with a KeyError is not the best way to handle a missing required config option imho
Not the best at all, but better than silently using `http://gitlab.renku.build` I would say.

> the app should fail already when helm tries to render the chart, but if deployed outside that and required config options are missing it should give a meaningful error and exit (in the cases where we can't provide a reasonable default).

Agree, but I think that's a bit outside the scope of this bugfix PR?

ableuler: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/462

jupyterhub/spawners.py
ableuler: Hmm, tbh instead of only taking the first part of the path, I'd rather leave it out completely and go with your original version. I just thought that adding the full path to the GitLab instance would complete what you had wanted to achieve

> I essentially decided to add all the parts of the repository url as separate labels

but when suggesting that I didn't think about the `/`. So if URL-encoding the path is not an option for you, I'd rather not add it.

jupyterhub/spawners.py
rokroskar: Sure that's fine - I wasn't sure why this PR was touching that code, but I didn't think it should be done like that. :)

jupyterhub/spawners.py
olevski: Ok I will remove the path label then.

jupyterhub/spawners.py
olevski: Ok so I will roll back all the changes I made for `os.environ["GITLAB_URL"]` so that they are back to using `.get` with a hardcoded default. Then we will fix this up in the issue that @ableuler created.

ableuler (approved):

PR #476, opened 2020-11-26
fix: remove git credentials

olevski:
This removes the git credentials from an image if the user is logged in but have less than maintainer permissions for the project.

Deployed currently at https://tasko.dev.renku.ch/.

The images that contains these changes can also be found at: 
- olevski90/git-clone:b71041c
- olevski90/renku-notebooks:b71041c


rokroskar:
cc @erbou

jupyterhub/spawners.py
```diff
@@ -238,6 +238,12 @@ def get_pod_manifest(self):
             working_dir=mount_path,
             security_context=client.V1SecurityContext(run_as_user=0),
         )
+        # if the user has valid git credentials and the access level of the user is admin or above
```
ableuler, line 241:
```suggestion
        # if the user has valid git credentials and the access level of the user is maintainer or above (ie the user can push to the repo)
```

ableuler (changes requested):

ableuler:
thanks for the quick action on this issue! 👍

jupyterhub/spawners.py
rokroskar: that's not maintainer afaik it would be developer, no?

rokroskar: or rather, it would really depend on the repos permissions - maybe you can use the gitlab API to figure it out?

olevski: So one of the differences between maintainer and developer is that a developer cannot push to protected branches and cannot create protected branches, while a maintainer can do that. A developer however can make regular branches and push to any non-protected branches.

So maybe it is better to make the check be if the user has developer access or above then keep the credentials. I just feel that restricting this to maintainer is a bit more conservative.

Here are the differences of all the access levels:
https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions

@rokroskar, @ableuler what do you think should I change that to developer or leave it as maintainer?

The access level check in `spawner.py` already finds the highest access level that the user has for the specific repo. And the access level values are as follows: https://docs.gitlab.com/ee/api/access_requests.html#valid-access-levels. The numbers indicate the level, 0: no access, 10: guest, 20: reporter, 30: developer, 40: maintainer, 50: owner (only for groups). So I already have this number for the user for the specific repo in the `self.gl_access_level` variable. We just have to decide what the cutoff is.

ableuler: 👍 for catching this @rokroskar. For me it clearly should be developer. Using maintainer her could be very confusing the for someone with developer access wanting to push to their own regular branch for example.

olevski: Ok I updated the code so that the cutoff for when the credentials are saved is developer or above. Also I revised the comment as suggested.

olevski:
Whoever is trying to test this out keep in mind that this is not deployed in my test environment anymore. I had to deploy something else. If you need it back let me know or deploy it yourself.

jupyterhub/spawners.py
rokroskar: actually what exactly a developer vs maintainer can do on protected branches is configurable:

image

But the point of the token is not to push to a protected branch, it's to push to any branch in the project which you can do as a developer, so I think that's the correct setting.

ableuler (approved):

pameladelgado:
> Whoever is trying to test this out keep in mind that this is not deployed in my test environment anymore. I had to deploy something else. If you need it back let me know or deploy it yourself.

Oh I wanted to test this, but I can use your images in my deployment to do it.

olevski:
@pameladelgado I deployed and tested anonymous sessions and it works.

pameladelgado (approved):
Thanks @olevski !
The code changes make sense.
I also tested this (as a maintainer, as a logged user with no push rights and as an anonymous user) and it works properly (also with a GitLab admin user).

PR #478, opened 2020-11-30
feat(app): move gitlab oauth token to separate proxy container

ableuler:
This PR adds a http proxy as sidecar container to the interactive notebooks. The users gitlab oauth token is not injected into the notebooks container anymore and can not be accessed from within that container. Instead, the git client is configured to use the http sidecar proxy which adds the necessary header for push/pull http request to/from the repo. The proxy only adds the authentication header to requests going to specified request URLs, per default this is only the git url of the repo for which the environment is launched.

Currently deployed on https://andreas.dev.renku.ch

What to improve / check:
- Verify if chosen approach makes sense.
- Check footprint of the chosen proxy, look for alternatives if too big. --> Seems ok on first sight, have to check when cloning a very big repo.
- ~Allow for multiple repo urls with potentially different authentication headers.~ I would do this in a later PR to not delay this.


erbou:
* Git pull/push and renku save/renku dataset ops works on a basic project, with LFS datasets,  and without git submodules
* Didn't test for performance penalty - looks ok at small scale, but true that it may affect network perfmances. Agree that it's a small cost to pay for the added security. We can aim for better perfs at a later time.
* Using sidecar containers (ambassador pattern?) for isolation makes a lot of sense in theory. Furthermore, it sounds like a good step toward interconnecting multiple protected github/gitlab instances to achieve a renku federation

aledegano (dismissed):

rokroskar (changes requested):
I tested this PR and noticed that the transfer rates for LFS were pretty severely affected. I could get only ~5 MB/s downloading data, whereas it was consistently > 50MB/s on dev.renku.ch. Maybe using another proxy would help? A traefik middleware comes to mind...

git-clone/entrypoint.sh
```diff
@@ -33,6 +33,13 @@ echo "https://oauth2:${GITLAB_OAUTH_TOKEN}@${GITLAB_HOST}" > .git/credentials
 git remote add origin $REPOSITORY
 git fetch origin
 
+git config http.proxy http://localhost:8080
+# Note: The proxy will still verify the certificates of
+#       the connection to the git server.
+git config http.sslVerify false
+git config --unset credential.helper
+rm .git/credentials
```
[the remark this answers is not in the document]
ableuler: At the moment the initial cloning doesn't go through the proxy since the init container has to complete before the other ones (including the sidecar proxy) are started, so didn't touch this logic for the moment and just made sure to clean up afterwards

ableuler:
> I tested this PR and noticed that the transfer rates for LFS were pretty severely affected. I could get only ~5 MB/s downloading data, whereas it was consistently > 50MB/s on dev.renku.ch. Maybe using another proxy would help? A traefik middleware comes to mind...

Hmm, that's bad! Thanks for testing. I will see if I can improve the settings of this `mitmproxy`, but it might be that we will need another proxy here. Traefik is not an options since it is a reverse proxy, but what we need here is a regular http proxy.

ableuler:
> I could get only ~5 MB/s downloading data, whereas it was consistently > 50MB/s on dev.renku.ch

@rokroskar would you mind trying the same setup again? I've deployed a solution with a node http proxy on andreas.dev.renku.ch

rokroskar:
This is better - I think it gets around 20 MB/s. On dev.renku.ch the same project gets about 25 MB/s. Both of these are super low rates and we should investigate why, but it doesn't look like the proxy is killing the transfer.

ableuler:
> This is better - I think it gets around 20 MB/s. On dev.renku.ch the same project gets about 25 MB/s. Both of these are super low rates and we should investigate why, but it doesn't look like the proxy is killing the transfer.

I just did another try and got 50 MB/s through the proxy, and then some random download from a test server that does not involve the proxy with only 5MB/s, so I agree with you that we have a problem here, but that the proxy is presumably not the bottleneck.

git-https-proxy/mitmproxy.js
```diff
@@ -0,0 +1,77 @@
...
+   * Request handler for the proxy. Checks if we're dealing with a
+   * push/pull to/from the repo, if yes add the users GitLab oauth
+   * token to the authorization header.
+   */
+
+  // A bit annoying that we have to reverse-engineer the request url here...
+  let requestUrl = (
+    `${ ctx.proxyToServerRequestOptions.agent.protocol }` +
+    `//${ ctx.clientToProxyRequest.headers.host }` +
+    `${ ctx.clientToProxyRequest.url }`
```
lorenzo-cavazzi, line 57: This could be simplified a bit by removing `${ ... }` wherever you don't need to include extra chars
```
    `${ ctx.proxyToServerRequestOptions.agent.protocol }//` +
    ctx.clientToProxyRequest.headers.host + ctx.clientToProxyRequest.url
```

-----

UPDATE: Ok nevermind the following comment, testing it seems to confirm that it works as expected :shrug: 
> Adding auth header to request: https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/old-private.git/info/refs?service=git-upload-pack

Is a `:` missing here or does `ctx.proxyToServerRequestOptions.agent.protocol` already includes the colon? I think this would produce `http//myhost[...]` but you probably need `http://myhost[...]`.

git-https-proxy/mitmproxy.js
```diff
@@ -0,0 +1,77 @@
...
+).toString('base64');
+const repoUrl = new url.URL(process.env.REPOSITORY_URL);
+
+let defaultPort;
+switch (repoUrl.protocol) {
+  case 'https:':
+    defaultPort = 443;
+  case 'http:':
+    defaultport = 80;
+}
```
lorenzo-cavazzi, line 36: Nothing wrong here, but I guess you want to always set a `default` case.
I would simplify a bit the syntax if `80` is also the default port:

```
const defaultPort = repoUrl.protocol === 'https' ?
  443 :
  80;
```

lorenzo-cavazzi (commented):
Looks good! Very nice addition.
I tried to deploy it in my environment and push/pull on both public and private repo and it seems to work well :+1: 
I added a couple of minor comments on the JS code.

git-https-proxy/mitmproxy.js
ableuler: Implemented the multiline template string suggestion, thanks! 👍 
The colon comes from with `ctx.proxyToServ.erRequestOptions.agent.protocol`

git-https-proxy/mitmproxy.js
ableuler: Good point - added the `default:` case. However, I prefer to keep it undefined if the protocol is not explicitly set to 'http' or 'https' (which should never happen, but still).

lorenzo-cavazzi (approved):
All good from my side :+1:

rokroskar (approved):

PR #482, opened 2020-12-10
chore: update JupyterHub

pameladelgado:
Updated JupyterHub and Python version in the hub.
This is deployed in https://pamela.dev.renku.ch/


aledegano (approved):

olevski (dismissed):
Pamela keep in mind that this is behind the master branch by a few commits. But I tested on my environment on a local branch after I updated with the latest from master and everything works. You should probably update this branch though before you merge.

ableuler:
Thanks for this update. As far as I can tell from looking at [the requirements of the base image we're using](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.10.6/images/hub/requirements.txt), the kubespawner and tmpauthenticator dependencies defined there have now caught up with the ones we've been installing on top. So I think we could remove those two lines from our `jupyterhub/requirements.txt`.

pameladelgado:
> Thanks for this update. As far as I can tell from looking at [the requirements of the base image we're using](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/0.10.6/images/hub/requirements.txt), the kubespawner and tmpauthenticator dependencies defined there have now caught up with the ones we've been installing on top. So I think we could remove those two lines from our `jupyterhub/requirements.txt`.

good point, I just removed those.

olevski (approved):
Sorry @pameladelgado I missed the notification about this.

PR #486, opened 2020-12-13
fix: correct order of lfs pull and repo config for git proxy

ableuler:
This also needed a bit of cleanup and refactoring of the autosave branch handling in the init container.


git-clone/entrypoint.sh
```diff
@@ -32,58 +32,63 @@ git config credential.helper "store --file=.git/credentials"
...
-  (git checkout ${BRANCH} || git checkout -b ${BRANCH})
-  git submodule init && git submodule update
-  git reset --hard $COMMIT_SHA
+    IFS='/' read -r -a AUTOSAVE_REMOTE_BRANCH_ITEMS <<< "$AUTOSAVE_REMOTE_BRANCH"
 
-  chown ${USER_ID}:${GROUP_ID} -Rc ${MOUNT_PATH}
-  exit 0
-fi
+    # Check if found autosave branch has valid format, fail otherwise
+    if [ "${#AUTOSAVE_REMOTE_BRANCH_ITEMS[@]}" -lt 7 ] ; then
```
olevski, line 64: @ableuler does this mean that if the `AUTOSAVE_REMOTE_BRANCH_ITEMS` array has less than 7 items then the branch is invalid and the autosave will not be recovered? Why the 7 threshold?

olevski (commented):
I tested this with a private project that has LFS files and things work. With autofetch on I have the files there right away without running the extra command to download files.

I just have one question about the code.

The thing about conflicting autosave branches with 7 character commit shas that match is not a problem IMO. Isn't the probability very low of getting the first 7 characters of the commit sha that are the same for two or even more branches?

git-clone/entrypoint.sh
ableuler: The full branch name is split at every occurrence of a `/`, and the resulting array must have 8 or more elements. Consider for example `remotes/origin/renku/autosave/andi/master/f329fe7/f329fe7`

ableuler:
> The thing about conflicting autosave branches with 7 character commit shas that match is not a problem IMO. Isn't the probability very low of getting the first 7 characters of the commit sha that are the same for two or even more branches?

The naming of the autosave branches is determined [here](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/e960e363a4c3ec6cee30cfeacde62f5662b64e89/helm-chart/renku-notebooks/templates/configmap.yaml#L56). So the commit SHA we're looking at the commented location is from the commit previous to the autosave commit. So basically the code here silently assumes that for one user/branch/commit-sha combo, there can be no more than one autosave branch. We kind of enforce this by always dropping the user into an existing autosave environment (and deleting the autosave branch while doing this) and by allowing only one running session per user/branch/commit, so one would have to almost know how to trick the system to end up with two autosave branches for the same base commit. Therefore I agree, my comment is probably not super relevant and should definitely not be phrased as a TODO. I'll change that.

ableuler:
> I tested this with a private project that has LFS files and things work. With autofetch on I have the files there right away without running the extra command to download files.

If you haven't done so already, I'd be very grateful if you could also test the autosave recovery. As far as I can tell it still works in the same way, but someone else touching it would be great.

olevski:
@ableuler I can confirm that uncommitted work is successfully recovered.

olevski (dismissed):

ableuler:
@olevski I improved the language of the comments a bit - can you please re-approve? Thx!

olevski (approved):

PR #493, opened 2020-12-21
bug: launching sessions that are part of a subgroup fails

### LINKED ISSUE #492: bug: private projects that are part of a subgroup or group fail to launch an session
This was reported by a user: https://renku.discourse.group/t/breaking-changes-recently/227

It is occurring because when we create the image pull secret for the private session we add a k8s label with the gitlab namespace. When the project belongs to a subgroup the project namespace is something like `group/subgroup`. '/' are not allowed in k8s label values and the creation of the secret as well as the whole session fails because of this.

Relevant part of the logs:
```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Secret \"xxxx-xxxxx-registry-db05d842-9369-447c-a5a1-a5343ffec86b\" is invalid: metadata.labels: Invalid value: \"xxx-xxx/xxx/xxxxx\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"
```

olevski:
closes #492 

I think we should discuss how to handle this exactly. But the reason this occurred is because when a project is part of a group the namespace is actually something like `group/subgroup`. This is then added to the pod and registry pull image secret k8s labels. However k8s does not accept `/` as a valid label value. That is why a user session could not be launched. 

Escaping these invalid characters is not an option IMO because then we have ambiguity with the labels. I.e. if we escape `/` with `-` then we cannot distinguish between someone naming a group like `this-is-a-group` and a group subgroup combo like `this-is/a-group`. 

I think the way to tackle this that includes the least amount changes is to convert the labels whose values are likely to contain invalid characters into annotations. Because such limits on characters are not applied to annotations by k8s. So I converted the `namespace` and `git-host` labels into annotations. I think that `git-host` is also likely to have `/` in its value depending on how gitlab was set up. The additional code in this PR is in the part that cleans up the old registry pull secrets and tries to find the pod where the secret was used based on the labels. The k8s API can filter itself based on label values but now that I have removed some labels used for this filtering I have to apply the filtering manually in python based on the labels that were converted to annotations.

Let me know what you think.


ableuler:
Thanks for proposing this fix. What you suggest makes a lot of sense to me. I have two questions:

1. I think this will introduce a change to the response of a GET request to the `/servers` endpoint. To what extent does the UI rely again on certain information being present as labels @lorenzo-cavazzi? 
2. Related to the above. Would it make sense to put all Renku specific information into annotations (at the cost of a breaking API change) and then simply duplicate the information we need for filtering k8s objects into labels, potentially even stripping these labels from before forwarding the API response in the notebook service? This might separate the role of the annotations and labels a bit more cleanly.

lorenzo-cavazzi:
We do use most of the information we receive from the `GET /servers` endpoint -- I believe we recently wrote a list in an issue/PR but I can't find it right now.
What is the breaking change here?

ableuler:
> What is the breaking change here?

Some information that is currently provided as pod label would be stored in a pod annotation instead.

olevski:
@ableuler @lorenzo-cavazzi there is no breaking change because I merge the labels and annotations in the list of servers that is returned to the API here: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/493/files#diff-178ed67454af02e29789bf4644237d084517fa656010addbf072377e255a2ed3R110

More specifically the merging of labels and annotations into the annotations is done here: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/493/files#diff-178ed67454af02e29789bf4644237d084517fa656010addbf072377e255a2ed3R194

So the the UI there is nothing missing from what is returned from a get request to servers or servers/<servername>

olevski:
@ableuler with regard to this suggestions see my comments below:

> Related to the above. Would it make sense to put all Renku specific information into annotations (at the cost of a breaking API change) and then simply duplicate the information we need for filtering k8s objects into labels, potentially even stripping these labels before forwarding the API response in the notebook service? This might separate the role of the annotations and labels a bit more cleanly and allow us to adapt the inner workings of how we retrieve objects from the k8s API without introducing API changes too frequently.

It is very easy to make sure that when we return stuff from the API the UI gets the same things as before. So I think whatever we decide to do we should do it without breaking changes with respect to what the UI expects from the API.

I like what you are proposing. Just to confirm the change would be like this:
- right now in this PR we have:
  - labels: username, commit-sha, projectName
  - annotations: namespace, projectId, branch, repository, git-host
- your proposal:
  - labels: username, commit-sha, projectName
  - annotations: namespace, projectId, branch, repository, git-host, username, commit-sha, projectName
  - when returning information from the `servers` or `servers/<server_name>` endpoints fully strip all labels
  - because all the labels are also duplicated as annotations there will be no breaking changes for the UI

p.s. When we return information about the active user servers from the API we use information only from the annotations - so keeping everything in the annotations and making copies as necessary in the labels should be the least disruptive too.

ableuler:
Thanks @olevski for the clarification. I was under the (wrong) assumption that we were returning the `server` objects more "as-is", including both labels and annotations and missed the part where you add the labels as annotations to the response.

> I like what you are proposing. Just to confirm the change would be like this:
> 
> * right now in this PR we have:
>   
>   * labels: username, commit-sha, projectName
>   * annotations: namespace, projectId, branch, repository, git-host
> * your proposal:
>   
>   * labels: username, commit-sha, projectName
>   * annotations: namespace, projectId, branch, repository, git-host, username, commit-sha, projectName
>   * when returning information from the `servers` or `servers/<server_name>` endpoints fully strip all labels
>   * because all the labels are also duplicated as annotations there will be no breaking changes for the UI

Yes, that's exactly what I had in mind. Because of my wrong assumption I thought that this would introduce a breaking API change, you're right that this is not the case (even better 😅 ).

olevski:
@ableuler I will make the changes you suggested.

olevski:
@ableuler I implemented and tested the changes we discussed. Let me know if you want me to change anything else.

ableuler (approved):
LGTM! 👍

PR #497, opened 2021-01-06
fix: properly recognize pod status when terminating

### LINKED ISSUE #496: bug: user pod status is not accurate
This problem is currently present on dev.renku.ch but not on renkulab.io. 

The problem is when you try to stop a pod the pod status goes to pending and then back to running, even though in reality the user pod is just taking a while to shut down. I think the main cause for this is this part of the code https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/util/kubernetes_.py#L152-L161 which seems to decide the status of the pod based on the first container. However with the addition of the https proxy container now there are two containers in each user pod and the status of each should be taken into consideration when the API sends this info to the UI.

This is not a hard fix at all. I will submit a PR today.

olevski:
closes #496 

Basically with the addition of the http proxy container in the pod the status of the pod cannot be determined just from one container. In addition even with a single container we had a problem where the "Terminating" status was not properly recognized. It seems that the only way that the "Terminating" status can be recognized is by checking whether `pod.metadata.deletionTimestamp` is None or not.

### DIFF
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -153,7 +153,11 @@ def _get_all_user_servers(user):
         ready = True
         try:
             for status in pod.status.container_statuses:
-                ready = ready and status.ready
+                ready = (
+                    ready
+                    and status.ready
+                    and pod.metadata.get("deletionTimestamp", None) is not None
+                )
         except (IndexError, TypeError):
             ready = False
 
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -156,7 +156,6 @@ def _get_all_user_servers(user):
                 ready = (
                     ready
                     and status.ready
-                    and getattr(pod.metadata, "deletion_timestamp", None) is None
                 )
         except (IndexError, TypeError):
             ready = False
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -156,7 +156,7 @@ def _get_all_user_servers(user):
                 ready = (
                     ready
                     and status.ready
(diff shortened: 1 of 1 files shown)


renku_notebooks/util/kubernetes_.py
```diff
@@ -150,8 +150,14 @@ def sort_conditions(conditions):
         return {"step": c.type, "message": c.message, "reason": c.reason}
 
     def get_pod_status(pod):
+        ready = True
```
ableuler, line 153:
```suggestion
        ready = getattr(pod.metadata, "deletion_timestamp", None) is None
```

renku_notebooks/util/kubernetes_.py
```diff
@@ -150,8 +150,14 @@ def sort_conditions(conditions):
...
         try:
-            ready = pod.status.container_statuses[0].ready
+            for status in pod.status.container_statuses:
+                ready = (
+                    ready
+                    and status.ready
+                    and getattr(pod.metadata, "deletion_timestamp", None) is None
```
ableuler, line 159:
```suggestion
```

ableuler: The pod metadata is the same for every container, so suggesting to take it out of the loop.

ableuler (changes requested):
So just to make sure I understand this correctly - while terminating the status of the containers can still be ready while the pod has a deletion timestamp? Would not have expected this... Anyway, thanks for fixing this. A behaviour that we've probably all seen before and generously ignored... 😅 .

olevski:
@ableuler thanks for the review. I accepted your suggestions and re-tested to make sure everything works.

ableuler (approved):

PR #511, opened 2021-01-15
fix: extraEnv requires only string values now

olevski:
So I saw the `helm lint` failed after the latest changes. But the line that causes the failure has not been changed in months.

I think this is occurring because jupyterhub was upgraded.

This is the line in the jupyterhub helm chart _helpers.tpl that causes [this problem](https://github.com/SwissDataScienceCenter/renku-notebooks/runs/1707349028):
```
{{- else if eq (typeOf $value) "string" -}}
- name: {{ $key | quote }}
  value: {{ $value | quote | println }}
{{- else }}
{{- printf "?.extraEnv.%s had an unexpected type (%s)" $key (typeOf $value) | fail }}
{{- end }}
```
It seems that before non-string values were allowed in the extraEnv part, but not anymore. The lint command passes after this.


pameladelgado (approved):
Good catch! :+1:

PR #513, opened 2021-01-15
fix: pod resources schema

olevski:
This will address the following errors:
- https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/4940/events/57bfd7f7151d4d5bb11d08825e054325/
- https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/4939/events/9c1dce46c5c94a8a9ee3511f367b2082/

The issues were occurring because:
- now we also have an `ephemeral-storage` variable that we did not know before in the pod resources
- the timestamp that when the pod starts can be `null` sometimes instead of a timestamp

### DIFF
--- a/renku_notebooks/api/schemas.py
+++ b/renku_notebooks/api/schemas.py
@@ -77,21 +77,15 @@ class UserPodAnnotations(
         return flatten_dict(data)
 
 
-class UserPodResources(
-    Schema.from_dict(
-        {
-            "cpu": fields.Str(required=True),
-            "memory": fields.Str(required=True),
-            "ephemeral-storage": fields.Str(required=False),
-        }
-    )
-):
-    """
-    Memory and CPU resources that should be present in the response to creating a
-    jupyterhub noteboooks server.
-    """
-
-    pass
+UserPodResources = Schema.from_dict(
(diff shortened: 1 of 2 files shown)


renku_notebooks/api/schemas.py
```diff
@@ -77,14 +77,21 @@ def unnest_keys(self, data, **kwargs):
         return flatten_dict(data)
 
 
-class UserPodResources(Schema):
+class UserPodResources(
```
ableuler, line 80: I think this is equivalent to simply doing
```
    UserPodResources = Schema.from_dict(
        {
            "cpu": fields.Str(required=True),
            "memory": fields.Str(required=True),
            "ephemeral-storage": fields.Str(required=False),
        }
    )
```

ableuler (changes requested):
A tiny bit of cosmetics... ;)

renku_notebooks/api/schemas.py
olevski: Yeah I think you are right and it looks way better. I will change it.

ableuler (approved):
Thanks a lot for the quick turn around on this! 👍

PR #518, opened 2021-01-18
fix: retry when creating notebook fails with 500

### LINKED ISSUE #365: creating server fails with 500 from JH
Sometimes the k8s api request from the jupyterhub spawner fails when creating a new notebook server, resulting in a 500 sent back to the notebooks service and ultimate failure. It would be nice if we could handle this by retrying. 

### Context:

Sentry Issue: [RENKU-NOTEBOOKS-5](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/91/?referrer=github_integration)

```
creating server test_2020_08_13_0921_21-3df4f418 failed with 500
```

olevski:
closes #365 

Not sure if I should retry more than once or add a period when I just wait before retrying.

I went for the simplest case of not waiting and just retrying once.

### DIFF
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -157,14 +157,14 @@ class SpawnerMixin:
 
         try:
             result = yield super().start(*args, **kwargs)
-        except ProtocolError as err:
+        except ProtocolError:
             self.log.warning(
                 "Spawning a JH server failed with ProtocolError, "
                 f"user_options {self.user_options}, args: {' '.join(self.get_args())}, "
                 f"auth_state has keys: {list(auth_state.keys())}, "
                 f"oauth_token is None: {oauth_token is None}"
             )
-            raise(err)
+            raise
         return result
 
 
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -24,6 +24,7 @@ import escapism
 import gitlab
 from kubernetes import client
 from tornado import gen, web
+from urllib3.exceptions import ProtocolError
 
 from kubespawner import KubeSpawner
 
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi:
Sorry for chiming in late, I notice only now the info you added to the reference issue.
It would be nice to further dig into the possible connectivity/networking issues, but I guess that would be a long process, possibly not leading to a solution.
Something I noticed is that the problem happens frequently if the user is inactive for some time. In that case, he gets the `500` error in the first trial but it works fine later. He doesn't get other `500` errors when starting other environments just after.
That makes me believe it could be somehow related to authentication (the first step contributes to refreshing the token?) but I didn't really try to go further in understanding this, so my conjecture could be totally wrong.

Anyway, the workaround you implemented looks fine. I would add a `sleep` before re-trying, 1 or 2 seconds should be enough.

olevski:
@lorenzo-cavazzi let my try to see if the 500 error shows up when I take the steps you describe.

ableuler (changes requested):
I like the fact that this logic is now on the notebooks-service side and not just in the client. I agree with @lorenzo-cavazzi that a bit of wait time before re-trying would be good, but I wouldn't add too much as the connection between the notebooks api any client will be kept open during this time. 
Also, it would be nice if you could catch this exception [in the spawner](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/4df28bc378b0c0c18fb1db37aac9224c682bcdbb/jupyterhub/spawners.py#L157), maybe add some sanitized context (either to the sentry event or simply to the JH logs) about the login status and the pod spec that we're POSTing to the k8s API and then re-raise the exception to have it tracked in sentry.

olevski:
@lorenzo-cavazzi I could not replicate the error by completely removing the methods that refresh a token. The only thing that happened is I was just asked to log back in.

Also as requested I added a 1 sec sleep before retrying to launch the user server when it fails the first time.

@ableuler I added a try statement that would capture the `ProtocolError` that happens in the spawner when this error occurs. Also I added a warning to the log that captures some of the variables in the spawner before re-raising. Let me know if this is sufficient. I did not want to log the token or any more of the contents of the auth_state dictionary.

Let me know if you want me to change anything.

ableuler (approved):

ableuler:
> I did not want to log the token or any more of the contents of the auth_state dictionary

That's exactly what I had in mind with "sanitized context" 👍

PR #529, opened 2021-01-27
feat: validate server options

### LINKED ISSUE #71: validate server_options
The `server_options` need validation in several places:

* in the `values.yaml` that the admin passes on deployment
* in the service code where the request with`serverOptions` in the body is processed

It's not obvious what the validation should consist of, however. 

1. Enforce only a specific set of server options, e.g. `resources.cpu_request`, `resources.mem_request` etc. 
2. Make sure that the `serverOptions` that are passed in through the request conform to the types specified in the `values.yaml`.
3. ?

This validation was left as a todo from pr #68.

olevski:
closes #71 

So I implemented evaluation on the helm chart values but only for the server options portion. I will add a new issue to add it for the whole values.yaml file.

I also added more detailed validation (in line with the values.yaml validation) for the API.

I did some digging online to see if marshmallow can parse jsonschema and convert to marshmallow type schemas and it cannot unfortunately. There is a library that can convert a marshmallow schema to jsonschema but it cannot include the custom validation we need for some of the server options stuff. So I think we unfortunately have to have this kind of duplication.

This is deployed at https://tasko.dev.renku.ch/

### DIFF
--- a/renku_notebooks/api/custom_fields.py
+++ b/renku_notebooks/api/custom_fields.py
@@ -1,6 +1,7 @@
 from typing import List, Mapping, Any
 from marshmallow import fields
 from marshmallow.exceptions import ValidationError
+import re
 
 
 class UnionField(fields.Field):
@@ -48,3 +49,13 @@ class UnionField(fields.Field):
                 errors.append(error.messages)
 
         raise ValidationError(errors)
+
+
+serverOptionCpuValue = fields.Number(
+    validate=lambda x: x > 0.0 and (x % 1 >= 0.001 or x % 1 == 0.0), required=True,
+)
+serverOptionMemoryValue = fields.String(
+    validate=lambda x: re.match(r"^(?:[1-9][0-9]*|[0-9]\.[0-9]*)[EPTGMK][i]{0,1}$", x)
+    is not None,
+    required=True,
+)
--- a/renku_notebooks/api/schemas.py
+++ b/renku_notebooks/api/schemas.py
@@ -2,7 +2,18 @@ from marshmallow import Schema, fields, post_load, post_dump
 import collections
 
 from .. import config
-from .custom_fields import UnionField
+from .custom_fields import (
+    serverOptionCpuValue,
+    serverOptionMemoryValue,
+)
+
+
+class LaunchNotebookRequestServerOptions(Schema):
+    defaultUrl = fields.String(required=True)
+    cpu_request = serverOptionCpuValue
+    mem_request = serverOptionMemoryValue
+    lfs_auto_fetch = fields.Bool(required=True)
+    gpu_request = fields.Integer(strict=True, validate=lambda x: x >= 0)
 
 
 class LaunchNotebookRequest(Schema):
@@ -14,9 +25,7 @@ class LaunchNotebookRequest(Schema):
     commit_sha = fields.Str(required=True)
     notebook = fields.Str(missing=None)
     image = fields.Str(missing=None)
-    server_options = fields.Dict(
-        keys=fields.Str(), missing={}, data_key="serverOptions"
-    )
+    server_options = fields.Nested(LaunchNotebookRequestServerOptions(), missing={})
 
 
 def flatten_dict(d, parent_key="", sep="."):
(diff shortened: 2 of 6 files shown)


olevski:
issue for adding validation for the whole values.yaml file #530

ableuler (changes requested):
Thanks for the careful work on validating that the server options conform to the desired schema. I agree that the duplication is a bit annoying, but we can live with it. 

The really important point still missing here (you might consider this simply a bug of the initial implementation of those server options) is an actual check if the server options which come as part as a POST body are elements of the options that were offered for selection in the first place. Basically, we have to prevent this

image

when the dialog only offers this

image.

I simply replayed the POST request from the browser while giving myself a bit more CPU for this.

renku_notebooks/api/custom_fields.py
```diff
@@ -48,3 +49,13 @@ def _deserialize(
                 errors.append(error.messages)
 
         raise ValidationError(errors)
+
+
+serverOptionCpuValue = fields.Number(
```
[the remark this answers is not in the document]
olevski: So the problem with these validations is as follows. Kubernetes reports the following as the regex used to validate them: 
`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`. However this lets 'through' some additional values that are then further checked after they pass the initial regex check. For example according to this regex a value of `-100M` is a valid for memory and a few other weird cases but these will not pass and a different error message will be shown.

To simplify things I decided to:
- just use regex for validation
- using regex only for validation allows me to easily encode this in the values.yaml validation and marshmallow, I can easily write fancy stuff in python for marshmallow but that is not the case for the helm values.yaml validation so I want to keep things simple
- with this regex capture the subset of valid options that look good in the UI and are easy to in the json-schema
- these 'look-good' valid options are then:
  - cpu: number greater than zero with at most three decimals (that is the the most precision k8s does for cpu)
  - memory: whole number followed by either E,P,T,G,M,K, or Ei,Pi,Ti,Gi,Mi,Ki
  - gpu: whole number equal to zero or greater
 
I think this still covers most cases. What do you think?

renku_notebooks/api/custom_fields.py
```diff
@@ -48,3 +49,13 @@ def _deserialize(
...
+    validate=lambda x: x > 0.0 and (x % 1 >= 0.001 or x % 1 == 0.0), required=True,
+)
+serverOptionMemoryValue = fields.String(
+    validate=lambda x: re.match(r"^(?:[1-9][0-9]*|[0-9]\.[0-9]*)[EPTGMK][i]{0,1}$", x)
```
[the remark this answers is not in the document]
olevski: k8s understands both. G is a power of 10 consistent with SI prefixes in general, while Gi is a power of 2. So for example 1K = 1000 bytes, 1Ki = 1024 bytes.

olevski:
@ableuler let me know what you think about my responses. Also I added validation for the server options in the post request for starting a server as you suggested. I am sorry I missed that. I also added a test to ensure that trying to start a server with bad options will fail.

renku_notebooks/api/custom_fields.py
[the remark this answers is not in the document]
ableuler: Ok, makes sense 👍

renku_notebooks/api/custom_fields.py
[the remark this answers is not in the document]
ableuler: oh true, forgot about this distinction! I think I'll never get used to the fact that computer scientists thought that it was convenient to denote 1024 simply as 1000...

renku_notebooks/api/schemas.py
```diff
@@ -1,8 +1,45 @@
...
+    cpu_request = serverOptionCpuValue
+    mem_request = serverOptionMemoryValue
+    lfs_auto_fetch = fields.Bool(required=True)
+    gpu_request = fields.Integer(strict=True, validate=lambda x: x >= 0)
+
+    @validates_schema
+    def validate_server_options(self, data, **kwargs):
+        server_options = read_server_options_file()
+        for option in data.keys():
+            if server_options.get(option, {}).get("type", None) == "boolean":
```
ableuler, line 30: This does the right thing, but it took me a moment to process the logic. How about something like this?
``
if option not in server_options:
  continue
if server_options[option]["type"] == "boolean":
  continue
if data[option] not in server_options[option]["options"]:
  raise ValidationError...
 ```

ableuler (changes requested):

renku_notebooks/api/schemas.py
olevski: Ok sure I will keep the `get` statements to avoid errors when a key is missing.

ableuler: Don't server options always have a type and non-boolean server options always have options? Wouldn't in this case uncaught KeyError be more helpful in detecting a problem here than silently using an empty dictionary or list?

olevski: I typed too soon. I think I can fully get away from the `get` statements with what you propose.

olevski: @ableuler you are right. The `get` statements are gone.

ableuler (approved):

olevski:
Ok great thanks. Also ran the acceptance tests locally with docker and they passed.

PR #535, opened 2021-02-01
chore: adapt ci/cd logic

### LINKED ISSUE #534: Adapt notebooks CI/CD pipeline
- We currently issue an automatically created PR to update the renku requirements after each successful PR merge. This should only happen on tag. 
- The CI deployments should run against `@master` of renku instead of `@development`.

ableuler:
- Update renku requirements only when a new version is released.
- Run ci deployements / acceptance tests against the main branch in renku.

closes #534.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -52,7 +52,7 @@ jobs:
         RENKU_TESTS_ENABLED: true
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
         renku_notebooks: "@${{ github.head_ref }}"
-        renku: "@development"
+        renku: "@master"
     - name: Check existing renkubot comment
       uses: peter-evans/find-comment@v1
       id: findcomment
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
   publish-chart:
     runs-on: ubuntu-latest
     needs: test-chart
-    if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
+    if: "startsWith(github.ref, 'refs/tags/')"
     steps:
     - uses: actions/checkout@v2
       with:


.github/workflows/cd.yml
```diff
@@ -0,0 +1,30 @@
+name: CD
+
+on: [release]
+
+jobs:
+  publish-chart:
+    runs-on: ubuntu-latest
+    needs: test-chart
```
rokroskar, line 8: you're missing this step - copy it over?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        fetch-depth: 0
+    - name: set up environment variables
+      run: |
+        echo "GIT_USER=Renku Bot" >> $GITHUB_ENV
+        echo "GIT_EMAIL=renku@datascience.ch" >> $GITHUB_ENV
+    - name: Push chart and images
+      uses: SwissDataScienceCenter/renku/actions/publish-chart@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.DEPLOY_TOKEN }}
```
rokroskar, line 21: should we make this an org-level secret?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,30 @@
...
+        DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+    - name: Wait for chart to get published
+      run: sleep 120
+    - name: Update component version
+      uses: SwissDataScienceCenter/renku/actions/update-component-version@master
```
rokroskar, line 27: The PR that is triggered by this action will need to be changed back to merge to `master`

ableuler:
oh wow, now you were too quick @rokroskar, sorry. Quickly changed the logic here after realising that the last step didn't build the chart and thus relies on the previous ones.

.github/workflows/cd.yml
ableuler: Yep - the corresponding issue is here: https://github.com/swissdatasciencecenter/renku/issues/1914

.github/workflows/cd.yml
ableuler: Yep, noticed this too and decided to remain with "push" but filter on `tags` only.

.github/workflows/cd.yml
ableuler: yep, seems like a good idea.

rokroskar (approved):

PR #544, opened 2021-02-10
chore: a bunch of dependency updates

ableuler:
One of them patches a vulnerability.

/deploy


Dockerfile
```diff
@@ -1,13 +1,11 @@
-FROM python:3.7-alpine
```
rokroskar, line 1: why this change?

olevski: @ableuler why are we changing the base image? Was this also a security vulnerability?

ableuler: Dependabot upgraded the cryptography package which doesn't play well with the alpine based image (we'd have to add a rust compiler). After switching to the debian based python image the package is pip-installable again. But I quickly wanted to try with the `-slim` base image and compare sizes, should have labeled it as WIP.

ableuler: https://github.com/pyca/cryptography/issues/5771

ableuler: The slim image is actually smaller than the resulting alpine image with the necessary installs.

rokroskar: double win 🥇

ableuler:
anybody feels like approving the PR? :)

olevski (approved):

rokroskar:
ah sorry, I was waiting for acceptance tests to finish.. and then it was lunchtime 👍 😄

PR #546, opened 2021-02-10
fix: allow and include unknown annotations in schema

olevski:
So the schema validation for the annotations seems to be a bit too strict.

This will allow unknown annotations (not mentioned in the schema) to not trigger an error and to be included in the response.

It should fix this issue.
https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/5156/?environment=dslab2021renkulab&project=4&referrer=alert_email


pameladelgado (approved):
I've tested this in a deployment that exhibit the problem, the issue seems to be fixed.
Thanks!

ableuler (approved):

PR #551, opened 2021-02-14
chore: add remote debugging with vscode

olevski:
So I heard Ralf did this on renku-python so I wanted to add it to renku-notebooks too. The nice thing is that this allows you to open a VS Code instance directly on the pod that is running the API without dealing with some of the proxying that telepresence does.

Let me know what you think. Lately I have not been able to get telepresence to work for some reason.

But if you want to stick with telepresence that is perfectly fine.

Here is the [PR](https://github.com/SwissDataScienceCenter/renku-python/pull/1801) that Ralf did on renku-python that I followed to get this to work.


renku_notebooks/.vscode/launch.json
```diff
@@ -0,0 +1,13 @@
...
+      "name": "Python: Attach",
+      "type": "python",
+      "request": "attach",
+      "connect": {
+        "host": "localhost",
+        "port": 5678
+      }
+    }
+    ]
+}
```
ableuler, line 13: @olevski you can configure your vscode to add a newline at the end of each file automatically.

ableuler:
Rebased on top of master and recreated lockfile due to some conflict.

PR #553, opened 2021-02-15
fix: labels values are too long for k8s

olevski:
So I think I came across something similar before but I noticed a bug on Friday through sentry ([link here](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/5210/?environment=renkulab&project=4&referrer=alert_email)) that is the result of a really long project name. We add the project name in the user pod labels but the labels in k8s cannot be longer than 63 characters so when someone with a long project name tries to launch a server it fails.

The way I resolved this is that in the labels I cut down the project name and username to a max of 63 characters. These are only used to filter the user pods when deleting the registry pull secrets. Then I also added the full project name and username in the pod annotations - which have no length limit. The annotations are then used to further filter the pods (that have already been filtered by the k8s api with the labels) when deleting old secrets and it ensures that there are no collisions because of the truncation in the labels.

This is deployed on https://tasko.dev.renku.ch/. I also tested with a project name that is 70 characters long and it works.

I did the acceptance tests locally and they passed after I increased the time for which the tests wait after deleting a project (in order to confirm the project was really deleted). The code here does not affect the deletion of a project at all - for some reason it is taking a bit longer than usual for the project to disappear from the UI after it is deleted in GitLab.

### DIFF
--- a/cull_secrets/clean_user_registry_secrets.py
+++ b/cull_secrets/clean_user_registry_secrets.py
@@ -33,14 +33,13 @@ from kubernetes.config.incluster_config import (
 POD_LABELS = [  # unlikely to have invalid characters, used to quickly filter
     "renku.io/commit-sha",
     "renku.io/username",
-    "renku.io/projectName",
+    "renku.io/projectId",
 ]
 
 POD_ANNOTATIONS = [  # most likely to have invalid characters (i.e. /, \, etc)
     "renku.io/git-host",
     "renku.io/namespace",
     "renku.io/username",
-    "renku.io/projectName",
 ]
 
 
--- a/cull_secrets/clean_user_registry_secrets.py
+++ b/cull_secrets/clean_user_registry_secrets.py
@@ -33,7 +33,7 @@ from kubernetes.config.incluster_config import (
 POD_LABELS = [  # unlikely to have invalid characters, used to quickly filter
     "renku.io/commit-sha",
     "renku.io/username",
-    "renku.io/projectId",
+    "renku.io/gitlabProjectId",
 ]
 
 POD_ANNOTATIONS = [  # most likely to have invalid characters (i.e. /, \, etc)
--- a/cull_secrets/clean_user_registry_secrets.py
+++ b/cull_secrets/clean_user_registry_secrets.py
@@ -39,6 +39,8 @@ POD_LABELS = [  # unlikely to have invalid characters, used to quickly filter
 POD_ANNOTATIONS = [  # most likely to have invalid characters (i.e. /, \, etc)
     "renku.io/git-host",
     "renku.io/namespace",
+    "renku.io/username",
+    "renku.io/projectName",
 ]
 
 
--- a/jupyterhub/spawners.py
+++ b/jupyterhub/spawners.py
@@ -308,7 +308,7 @@ class RenkuKubeSpawner(SpawnerMixin, KubeSpawner):
         self.extra_annotations = {
             RENKU_ANNOTATION_PREFIX + "namespace": options.get("namespace"),
             RENKU_ANNOTATION_PREFIX
-            + "projectId": "{}".format(options.get("project_id")),
+            + "gitlabProjectId": "{}".format(options.get("project_id")),
             RENKU_ANNOTATION_PREFIX + "branch": options.get("branch"),
             RENKU_ANNOTATION_PREFIX + "repository": repository_url,
             RENKU_ANNOTATION_PREFIX + "git-host": git_host,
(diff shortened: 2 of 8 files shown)


pameladelgado (commented):
In principle this looks like a good work-around. It's important keep the full length for the pod label matching for the secrets cleanup like you did.

What would happen if you have the same user with two projects that have the same 63 first characters as project name? I guess this is a corner case, but would it be able to launch environments properly?

Couldn't we use the project's GitLab ID? that should be unique enough...

olevski:
> What would happen if you have the same user with two projects that have the same 63 first characters as project name?

@pameladelgado this is a good point, but it is not a problem. The renku notebooks api reports the list of active user servers based the label selector for the username and heritage only (i.e. `f"heritage=jupyterhub,renku.io/username={safe_username}"`) and the rest of the filtering is done with the annotations. And the annotations do not have truncated values for anything. So the filtering and everything works.

I tested this out on my deployment with these two projects which have the same first 66 characters:
image
image

Everything works as expected.

olevski:
> Couldn't we use the project's GitLab ID? that should be unique enough...

@pameladelgado yes that is correct I will change it. It is better to avoid the truncation whenever we can.

olevski:
ok redeployed and reran the acceptance tests locally after changing the label to the project id - everything works.

pameladelgado:
oh sorry I didnt see your earlier response.
Thanks, I agree that it's better to use the unique ID than to truncate the name. 
I noticed that we are still truncating the username, for this there's the GitLab user ID in case you want to also change this.

jupyterhub/spawners.py
```diff
@@ -318,10 +318,10 @@ def get_pod_manifest(self):
         }
         # some annotations are repeated as labels so that the k8s api can filter resources
         self.extra_labels = {
-            RENKU_ANNOTATION_PREFIX + "username": safe_username,
+            RENKU_ANNOTATION_PREFIX + "username": safe_username[:63],
             RENKU_ANNOTATION_PREFIX + "commit-sha": options.get("commit_sha"),
-            RENKU_ANNOTATION_PREFIX + "projectName": options.get("project"),
-            "hub.jupyter.org/network-access-hub": "true"
```
ableuler, line 324: This line is necessary for the user session pod to be able to communicate with the hub after network policies were enabled by default in the hub.

jupyterhub/spawners.py
```diff
@@ -318,10 +318,10 @@ def get_pod_manifest(self):
         }
         # some annotations are repeated as labels so that the k8s api can filter resources
         self.extra_labels = {
-            RENKU_ANNOTATION_PREFIX + "username": safe_username,
+            RENKU_ANNOTATION_PREFIX + "username": safe_username[:63],
             RENKU_ANNOTATION_PREFIX + "commit-sha": options.get("commit_sha"),
-            RENKU_ANNOTATION_PREFIX + "projectName": options.get("project"),
```
ableuler, line 323: Should we keep the project name as an annotation? I understand why you switched to the ID for filtering, but having the name in the annotations could still be useful for us when debugging, etc.

cull_secrets/clean_user_registry_secrets.py
```diff
@@ -33,12 +33,13 @@
 POD_LABELS = [  # unlikely to have invalid characters, used to quickly filter
     "renku.io/commit-sha",
     "renku.io/username",
-    "renku.io/projectName",
+    "renku.io/projectId",
```
ableuler, line 36: Maybe we can use gitlabProjectId here (and in other corresponding locations in the code) just to be extra clear?

ableuler (changes requested):
Thanks a lot for identifying this one! Having someone actually go after the sentry errors and hunt down the bugs is really great! 🎖️

jupyterhub/spawners.py
olevski: I am sorry I will put that back in. I remember seeing the PR that added that line. Not sure why I deleted it here. Good catch.

jupyterhub/spawners.py
olevski: This is the list of labels @ableuler. The project name already exists in the annotations. I did not remove that.

cull_secrets/clean_user_registry_secrets.py
olevski: Sure I will change the name.

jupyterhub/spawners.py
ableuler: ah sorry, forgot for a moment that we duplicate all labels into annotations already anyway.

olevski:
Ok I addressed the comments from @ableuler, kept the username in the labels and added a check for the username length when a notebook is launched as we discussed.

I also ran the acceptance tests locally and they passed. This is also currently deployed at https://tasko.dev.renku.ch/

pameladelgado (approved):
Looks good to me!

ableuler (approved):

PR #557, opened 2021-02-16
fix: image pull secret missing for pod restart

### LINKED ISSUE #556: fix: image pull secret for private projects should remain as long as the user pod is present
It seems a user had an issue where her pod was not able to restart after being killed for consuming lots of memory. The pod could not restart because it could not pull the image. 

Here is a snippet of the log and the pod status reported `OOMKilled`
```
[I 2021-02-16 00:45:21.494 SingleUserNotebookApp restarter:110] KernelRestarter: restarting kernel (1/5), keep random ports
kernel 9e944106-95ec-47bf-abd0-e10d774415d0 restarted
/opt/conda/lib/python3.7/site-packages/jupyter_client/manager.py:358: FutureWarning: Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).
FutureWarning)
[I 2021-02-16 00:45:22.793 SingleUserNotebookApp restarter:110] KernelRestarter: restarting kernel (1/5), keep random ports
kernel ae3f7f0b-ebff-4c6d-9586-40f2ebf12e3e restarted
/entrypoint.sh: line 26: 24 Killed $@
```

I was able to reproduce very similar behaviour by:
- starting a server from a private project
- deleting the image pull registry secret when the server is fully started
- running some python code to consume all requested memory and getting the python process killed
- then shutting down the notebook container by running `kill 1` in its shell
- after all this the pod would be stuck in pending mode and the user sessions would not be rechable at all as the notebook container is trying to start back up but it cannot pull the private image since the image pull secret is gone

See here for the discussion: https://swiss-data-science.slack.com/archives/G016BR3Q9KJ/p1613465881004200

Also below is a summary of all the different behaviours when the image pull secret is missing or not.
image

To resolve this we will delete the image pull secret only if the user pod is not around at all. Currently we delete the secret as soon as the pod is fully started.

olevski:
closes #556 

Depends on and should be merged only after PR #553 is merged.

### DIFF
--- a/cull_secrets/clean_user_registry_secrets.py
+++ b/cull_secrets/clean_user_registry_secrets.py
@@ -33,14 +33,13 @@ from kubernetes.config.incluster_config import (
 POD_LABELS = [  # unlikely to have invalid characters, used to quickly filter
     "renku.io/commit-sha",
     "renku.io/username",
-    "renku.io/projectName",
+    "renku.io/gitlabProjectId",
 ]
 
 POD_ANNOTATIONS = [  # most likely to have invalid characters (i.e. /, \, etc)
     "renku.io/git-host",
     "renku.io/namespace",
     "renku.io/username",
-    "renku.io/projectName",
 ]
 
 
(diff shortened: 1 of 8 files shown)


ableuler (dismissed):
LGTM, happy to re-approve once the dependant PR is merged! :)

pameladelgado (approved):
Thanks for making this fix so quickly!
For me it's okay to merge this after #553 is.

olevski:
Can I get a re-approve here? I updated the branch accordingly and resolved conflicts after merging the dependent PR. Also I ran the acceptance tests locally and they passed. And if you are curious this is deployed at https://tasko.dev.renku.ch/.

ableuler (approved):

Issue #106, opened 2019-02-24
do not use local node disk space for user pods


rokroskar:
Two possibilities here: 

1. use PVs with retain policy `DELETE` so that the volumes are automatically removed when the pod is gone. 
2. have a large shared volume where we keep state of user sessions - this would require a service to do maintenance and cleanup of these directories. 

The first option above should be easy to implement. It involves the k8s admin creating a `temporary` (or equivalent name) `storageClass` and changing the volume spec in the spawners code. We should check if this `storageClass` exists before trying to use it and if not use `emptyDir` as we do now.

----

_(edited by @ableuler 2020-11-18)_
I would like to add two more options:

3. use PV's with retain policy `DELETE` similar to 1), but make the user environments `statefulSets` rather than "naked" pods. This would allow a pod to be recreated (for example in the case of a node failure) and prevent users from loosing data.

4. rather than having persistent state of individual user sessions, we could have persistent space for users, similar to a users home directory on shared HPC infrastructure. A repo would simply be cloned into this space when an environment is launched. However, to enable this we would have to mount the same volume to different user pods (same as for 2). This is similar to what was requested in https://github.com/SwissDataScienceCenter/renku-notebooks/issues/149.

rokroskar:
We should more clearly outline the expected use-cases - for example, we might even have different "clases" of users (institutional vs. external) for which different resource limits should apply. For now we will enforce a 20GB limit for interactive sessions and work on setting up policies. Therefore this issue is pushed into the backlog.

ableuler:
The two options outlined in the description have very different goals in mind: 

1. Nothing changes for the user, except that they don't loose their work when someone pulls the plug on the node that is running their session.
2. There is some degree of persistence between different user sessions.

I would propose to start with option 1). It seems much clearer in terms of user story and I think having done 1) will also help us to hash out 2) at some point in the future. 

Option 1) could be implemented by modelling the user sessions as statefulSets which have a persistent volume attached. What is very unclear to me is to what extent the existing JupyterHub based way of spawning and exposing user sessions would have to be adapted for this.

ableuler:
Putting on pause for one sprint as other topics have higher priority.

PR #573, opened 2021-03-09
feat: use pvs for user sessions

### LINKED ISSUE #464: Use dynamically created PVs for interactive sessions
We create a dynamically provisioned PV (with its default retention policy `delete`) for each interactive session. The POST request to the `/servers` endpoint should accept a disk size parameter and the `GET` request to the `/server-options` should give a range or list of allowed sizes.

Open questions to be clarified and synced with the other components are:

- how does the user set the size of the volume?
- what are the allowed sizes, do we retain the volume for a period of time or is it immediately deleted when the session closes?
- improve/write documentation about managing data usage

### Update 4/2/2021:

following a design meeting, we have decided to implement the following: 

* allow the user to pick a volume size similar to the way memory and cpu are specified via a slider in the UI
* allow the user to set a default for the project in the project's configuration options
* allow the volume request in the user options to exceed the maximum value of the slider, with a hard-limit implemented and configured via the notebooks service

### LINKED ISSUE #466: Implement fault-tolerant user sessions
Based on the outcomes of https://github.com/SwissDataScienceCenter/renku-notebooks/issues/465, we should move forward and change the implementation of interactive sessions such that users don't loose their work when the node running there session goes down.

rokroskar:
This PR closes #464 and addresses #106 by providing a PV for each running interactive session. The change touches on many different aspects of interactive sessions functionality. 

* the PVCs use the "temporary" storage class which uses the `DELETE` retention policy - this means that once the PVC is deleted, the underying PV itself is also deleted
* the PV is only deleted if the session is terminated via the renku UI - otherwise the PVC remains untouched - this includes if the pod is removed by jupyterhub due to inactivity
* due to the fact that PVs are preserved in the case of an unwanted session termination, autosave branches are disabled if PVs are turned on - when the session starts, we check if a PVC for this user/project/commit already exists and if so we reuse it. If it does not we create a new one.
* a new server launch option "disk size" is introduced - this can be set by the instance admin like all of the other options

Note that a cluster role for PV creation has to be created for the service account - it is currently added to the helm templates, which means that if the deployment with PVs is attempted by a non-admin user it will fail. 

This PR also closes #466 by providing PVs that will survive a node crash and be automatically reused when necessary. 

Two things to note:

1. this PR will probably need to be followed up by some mechanism to clean up zombie PVs, i.e. ones that a user created once and never came back to. We should decide what the appropriate policy for this would be. 
2. The auto save branch functionality is completely disabled by the changes here, if PVs are used. This means that if something potentially goes wrong with the PV itself, the unsaved work will be lost. IMHO this is a small price to pay for avoiding the verbosity of autosave branches, but perhaps a middle-ground is possible. 

This is currently deployed in https://rok.dev.renku.ch - because it requires a cluster role (see above) and a value change, I have not added it as a default to CI yet.


jupyterhub/spawners.py
```diff
@@ -259,20 +268,21 @@ def get_pod_manifest(self):
...
-                "exec": {
-                    "command": [
-                        "/bin/sh",
-                        "-c",
-                        "/usr/local/bin/pre-stop.sh",
-                        "||",
-                        "true",
-                    ]
+        # 5. Configure autosaving script execution hook if we are not using a persistent volume
+        if not options.get("pvc_name"):
```
lorenzo-cavazzi, line 272: Supporting both approaches (PV and autosave) may require a few other changes. Currently, the UI fetches the autosave branches directly from GitLab.

We could add an endpoint to provide the information about an existing volume for a specific commit, or we can move all the autosave/volumes management to renku-notebooks. That requires a new endpoint to provide a list of the commits having an autosave for a target project.

I prefer this second option because it's cleaner, it masks unnecessary deployment details (branches vs PV), and it would make it simple to add another option in the UI to discard the previously saved work when starting a new pod.

renku_notebooks/api/notebooks.py
```diff
@@ -337,6 +360,20 @@ def stop_server(user, forced, server_name):
         return make_response(jsonify({"messages": {"error": "Server not found."}}), 404)
 
     r = delete_named_server(user, server_name)
+
+    # If the server was deleted gracefully, remove the PVC if it exists
+    if r.status_code < 300:
```
lorenzo-cavazzi, line 365: Would it make sense to allow the user to decide between "stop" and "hibernate"? In the second case, the PVS is kept when gracefully deleting.

renku_notebooks/util/kubernetes_.py
```diff
@@ -299,3 +300,77 @@ def _secret_exists(name, namespace):
...
+    git_host,
+    storage_size,
+    storage_class="default",
+):
+    """Create a PVC."""
+
+    # check if we already have this PVC
+    pvc = _get_pvc(username, project_id, commit_sha)
+
+    if pvc:
```
lorenzo-cavazzi, line 320: Should we also check for the size and adjust it if the user required more GB?

lorenzo-cavazzi (commented):
Great to see this in action!
It generally works well. There are 2 things possibly confusing for the user:
- The size requested by the user is ignored if a previous PV already exists. This is generally not a big deal if the user requested a smaller disk, but it may be annoying for the user trying to get a bigger disk.
- We should notify the user about the existence of a PV associated with a commit before starting the pod. Otherwise, it may be confusing to find uncommitted code.

I added a couple of other comments inline.

The code looks good to me. It would be nice if a python expert would also give it a look.

jupyterhub/spawners.py
rokroskar: so the API would return generically "there is some way of recovering unsaved work for these commits" regardless of whether it's an autosave branch or an existing PV?

renku_notebooks/api/notebooks.py
rokroskar: Hmm maybe - but I would see that as an extension of the present functionality. There was also some discussion to potentially provide long-lived volumes for users in general.

renku_notebooks/util/kubernetes_.py
rokroskar: Right, I also thought that might be confusing - it should be possible to resize the volumes afaik, I will try.

rokroskar:
Thanks for poking around the PR @lorenzo-cavazzi! I totally agree with the two points you raised. I think the resizing should be possible to do per default - the question is if we should resize or just inform the user what the size is and that they can't change it? The second suggestion is maybe a bit more involved - I also prefer the second option but I'm wondering if it would be fine to do it as a follow-up PR or to include it here.

jupyterhub/spawners.py
lorenzo-cavazzi: Exactly.

lorenzo-cavazzi:
I guess we can even skip the notification if the volume is bigger than the required size while expanding a smaller volume should probably be part of the PR (if it's reasonably easy).
We can skip the second part for now and make a follow-up issue, but I would tackle it soon. Users complained in the past when we introduced the autosave and we automatically restored the content without any notification. That would be seen as a small regression by those users.

renku_notebooks/util/kubernetes_.py
```diff
@@ -299,3 +300,77 @@ def _secret_exists(name, namespace):
...
+        status = "existing"
+    if not pvc:
+        pvc = client.V1PersistentVolumeClaim(
+            metadata=client.V1ObjectMeta(
```
ableuler, line 324: Can you add the branch name here? It would be nice to have all arguments of `make_server_name(namespace, project, branch, commit_sha)` easily available in order to identify a user server (or create the UserServer object once that refactoring is through) from the info that is easily available without any further API requests.

renku_notebooks/config.py
```diff
@@ -56,6 +56,12 @@
 """The default image to use for an interactive session if the image tied to the
 current commit cannot be found."""
 
+NOTEBOOKS_USE_PERSISTENT_VOLUMES = os.environ.get(
```
ableuler, line 59: I might have missed this, but it seems to me that this config is not used in the server_options endpoint, so could it be that if this is set to false we're advertising the disk option to the UI anyway?

ableuler (changes requested):
Looks really nice - I didn't get to test this yet, but the proposed changes make sense.

renku_notebooks/config.py
rokroskar: good point, should filter it out there. otoh we may want to impose limits on ephemeral volumes as well...

renku_notebooks/util/kubernetes_.py
rokroskar: I could just use `make_server_name` for the pvc name?

renku_notebooks/util/kubernetes_.py
```diff
@@ -299,3 +301,96 @@ def _secret_exists(name, namespace):
...
+    branch,
+    commit_sha,
+    git_host,
+    storage_size,
+    storage_class="default",
+):
+    """Create a PVC."""
+
+    # check if we already have this PVC
+    pvc = _get_pvc(username, project_id, commit_sha)
```
olevski, line 321: @rokroskar I would use the `servername` / `make_servername` here. The servername we pass on to JH and that is what JH uses to distinguish between different sessions. So it is nice to reinforce and reuse that logic with the pvc rather than filtering by labels. Then when we change `make_servername` we update the logic of what distinguishes unique servers and pvcs in one go.

If you use `servername` here you dont even have to filter by labels - you can just get the exact pvc - if it exists.

renku_notebooks/util/kubernetes_.py
olevski: @rokroskar ☝️ I think using `make_server_name` in the pvc name is a great idea. See my other comment about `_get_pvc` for my reasoning why.

renku_notebooks/api/notebooks.py
```diff
@@ -225,6 +227,25 @@ def launch_notebook(
...
+            name=pvc_name,
+            username=safe_username,
+            git_namespace=namespace,
+            project_id=gl_project.id,
+            project=project,
+            branch=branch,
+            commit_sha=commit_sha,
+            git_host=urlparse(config.GITLAB_URL).netloc,
+            storage_size=server_options.get("disk_request"),
+            storage_class="temporary",
```
olevski, line 245: So the `temporary` storage class is not available in different providers/k8s clusters. For example `gke` will have only `premium-rwo`, `standard` and `standard-rwo` by default. I had this issue when I was deploying renku on gcp. Jena uses also a `temporary` storage class (but this is adjustable in the `values.yaml` file for renku in the Jena/kg section). So I think that this value should also be specified by the user in the `values.yaml` file and not be hardcoded in here.

olevski: Also it may be that someone wants to use a fancier storage class backed by ssd or something else for the user sessions.

renku_notebooks/util/file_size.py
```diff
@@ -0,0 +1,58 @@
...
+    "gi": 1024 ** 3,
+    "ti": 1024 ** 4,
+    "pi": 1024 ** 5,
+    "ei": 1024 ** 6,
+    "zi": 1024 ** 7,
+    "yi": 1024 ** 8,
+}
+
+
+def parse_file_size(size_str):
```
olevski, line 46: I am ok to do this in a separate PR. But I think it would look much nicer if we make a custom marshmallow field for memory/disk and give it a `bytes` property which will do this calc. This way the logic for validating the field and for extracting a number is all in one place. Also we could even potentially "deserialize" the string value to an integer that is just bytes - this can be done by overriding the `_deserialize` method.

We can do this by replacing this:
```
serverOptionDiskValue = fields.String(
    validate=lambda x: re.match(r"^(?:[1-9][0-9]*|[0-9]\.[0-9]*)[EPTGMK][i]{0,1}$", x)
    is not None,
    required=False,
)
```

By something like this:
```
class InformationAmount(fields.String):
    def _validate(self, value):
        return re.match(r"^(?:[1-9][0-9]*|[0-9]\.[0-9]*)[EPTGMK][i]{0,1}$", value) is not None

    @property
    def bytes(self):
         # insert logic from parse_file_size
```

olevski (changes requested):
@rokroskar I just have a few very picky things. Also some of them can be addressed in separate PRs I think. Take a look.

renku_notebooks/util/kubernetes_.py
```diff
@@ -299,3 +301,96 @@ def _secret_exists(name, namespace):
...
+def _get_pvc(username, project_id, commit_sha):
+    """Fetch the PVC for the given username, project, commit combination."""
+
+    pvcs = v1.list_namespaced_persistent_volume_claim(
+        kubernetes_namespace,
+        label_selector=f"renku.io/commit-sha={commit_sha}, "
+        f"renku.io/gitlabProjectId={project_id}, "
+        f"renku.io/username={username}",
+    )
+    if pvcs.items:
```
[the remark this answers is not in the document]
rokroskar: good point, thanks! Do you throw an error in that case?

renku_notebooks/util/kubernetes_.py
rokroskar: good idea - I switched to using `make_servername` for the name itself but I didn't think to just fetch by name.

renku_notebooks/api/notebooks.py
rokroskar: definitely - it was on my todo to make this configurable but I forgot. I guess it will be up to the operator then to figure out how to actually remove the PVs? We could include a template for the "temporary" storage class in the chart so it's created by default if possible?

renku_notebooks/util/file_size.py
rokroskar: That's an excellent point - I was actually thinking how I should validate the string 🤦

renku_notebooks/util/kubernetes_.py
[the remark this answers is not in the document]
rokroskar: I've fixed this by simplifying this logic and just dealing directly with pvc names rather than fetching a list and filtering.

renku_notebooks/util/file_size.py
rokroskar: I'm not 100% sure where to go with this tbh, so I'll make an issue for it.

rokroskar: Follow up: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/582

rokroskar:
Thanks @olevski for the review! I think I've addressed most of it - I've made a follow up issue for the file size refactoring and another one for tests (https://github.com/SwissDataScienceCenter/renku-notebooks/issues/583).

jupyterhub/spawners.py
rokroskar: I haven't addressed this in this PR because I think it needs some API design - but I opened an issue and I suggest we discuss the design and tackle next sprint: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/584

olevski:
A few other things I saw with playing around with this:
- Should we update the script that makes the values.yaml file for developers? If we make the default here for pvs to be used then devs will need admin privileges to deploy this.
- Should we update the script in the renku repo that makes a sample of a values.yaml file? I guess this is optional since the default is not using pvs for user sessions?
- If I have unsaved work and then close my session with the UI, there is no backup and the work is fully gone. This does not occur when not using pvs because a branch to back things up is made in this case - but with the pvs this does not happen anymore.

olevski:
@rokroskar I am ok to approve the PR without resolving the things I mentioned above right now. I can make follow up issues for them. 

One final thing we discussed is the validation of the disk space request and the ability to bypass it by making a direct api call. However this is currently not the case. This [line](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/api/schemas.py#L39) in the code currently prevents accepting any server option request that does not validate against the server options file. So if we want to make this possible we need to edit the code in the line I linked to allow this. Let me know what you think.

olevski (approved):

renku_notebooks/api/schemas.py
```diff
@@ -37,6 +40,16 @@ def validate_server_options(self, data, **kwargs):
...
+                # check if we allow arbitrary options
+                if server_options[option].get("allow_any_value"):
+                    value_range = server_options[option].get("value_range")
+                    if not value_range:
+                        raise RuntimeError("You must specify a value range.")
+                    if not _in_range(data[option], value_range):
+                        raise ValidationError(
+                            f"The value {data[option]} is outside of the range "
+                            f"{value_range.get('min')} to {value_range.get('max')}."
+                        )
```
olevski, line 52:
```suggestion
                        )
                    return
```

olevski: I think I got the indent right. Without this return statement the code continues to hit the validation error below even thought it should not.

olevski: that should actually be `continue` - my bad

olevski (dismissed):

helm-chart/renku-notebooks/templates/clusterrole.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if and .Values.rbac.create .Values.userSessionPersistentVolumes.enabled -}}
```
ableuler, line 1: Should we try to start giving more indicative names to the templates? We now have 

- role.yaml
- rolebinding.yaml
- clusterrole.yaml (which containts the clusterrolebinding)

I already see myself being confused about where to find what. How about:

- /templates/rbac/image-pull-secrets.yaml
- /templates/rbac/session-pvc.yaml

both containing the (cluster)role and -binding? Maybe also name the actual k8s resources accordingly?

ableuler (changes requested):
This is great. Just a small comment about naming things (which we all know is one of the two hard things in software development 🤷 )

helm-chart/renku-notebooks/templates/clusterrole.yaml
rokroskar: thanks @ableuler - I've made a template called `session-pv-rbac.yaml` that includes a custom role/rolebinding and the cluster role/rolebinding.

olevski:
@rokroskar I ran the acceptance tests on this locally through docker. They all passed except the one that usually fails locally which is related to seeing and making isses/PRs in the renku ui. When I test this functionality manually everything works though.

The last thing that is outstanding is Anreas' comment above though. If you address that I can retest and approve.

```
[info] ScalaTest
[info] Run completed in 30 minutes, 50 seconds.
[info] Total number of tests run: 11
[info] Suites: completed 10, aborted 0
[info] Tests: succeeded 10, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed: Total 11, Failed 1, Errors 0, Passed 10
[error] Failed tests:
[error]         ch.renku.acceptancetests.MergeRequestsAndIssuesSpec
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 1899 s (31:39), completed Mar 22, 2021, 3:16:14 PM
```

rokroskar:
Thanks a lot @olevski!! Regarding the comment from @ableuler - I consolidated the rbac needed for PVs into a single file [here](https://github.com/SwissDataScienceCenter/renku-notebooks/pull/573/commits/bd48c70237d3ae437467c1e319fd9e9e7824f7e1).

olevski (approved):

ableuler (approved):
Thanks!

PR #603, opened 2021-04-13
fix: increase huboauth caching from 1min to 5min

olevski:
According to profiles run on the /servers endpoint, the execution time is mainly spent on two functions in the user class code: `_get_hub_username` and `_get_hub_user`. The _get_hub_username function just uses the `user_for_token` method of HubOAuth from jupyterhub. And if we adjust the caching that is provided by jupyterhub in HubOAuth then we can save about ~150-300 ms. These are two profiles for the call to the same endpoint by the same user. I assume that in the second one the cache for HubOauth was used and the execution is roughly halved.

The default max cache age for HubOAuth from Jupyterhub is 300 sec (https://jupyterhub.readthedocs.io/en/stable/api/services.auth.html?highlight=huboauth#jupyterhub.services.auth.HubOAuth.cache_max_age)

Long execution time profile
image

Short execution time profile
image

### DIFF
--- a/renku_notebooks/config.py
+++ b/renku_notebooks/config.py
@@ -33,7 +33,7 @@ JUPYTERHUB_API_TOKEN = os.environ.get("JUPYTERHUB_API_TOKEN", "")
 """The service api token."""
 
 JUPYTERHUB_ADMIN_AUTH = HubOAuth(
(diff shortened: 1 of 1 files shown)


ableuler (approved):
Seems reasonable for our use case.

PR #605, opened 2021-04-13
fix: unset lfs auth mode in init container

ableuler:
This PR fixes a problem which occurrs when two conditions were met simultaneously:

1. The git lfs auto fetch option is set to true on session start
2. There are files in LFS.

In this case, git lfs adds a line like `lfs.https://renkulab.io/gitlab/andreas/test-lfs-problem.git/info/lfs.access=basic` to the repos `.git/config` when pulling the files during the init container. Because of this configuration, git prompts for username and password later when a user tries to push from the main session container. Removing this config at the end of the init container entry point delegates authentication back to the git http proxy.

/deploy

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -96,7 +96,7 @@ jobs:
         RENKU_RELEASE: renku-ci-nb-${{ github.event.number }}
       run: |
         echo "$RENKUBOT_KUBECONFIG" > ${{ github.workspace }}/renkubot-kube.config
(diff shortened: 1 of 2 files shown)


rokroskar (approved):

olevski (approved):
Tested the described scenario and confirmed it occurs. After deploying this change the problem is indeed gone.

It is impressive you were able to track this down!

ableuler:
> Tested the described scenario and confirmed it occurs. After deploying this change the problem is indeed gone.

Thanks for testing this thoroughly!

PR #607, opened 2021-04-15
feat: endpoint for autosave information

### LINKED ISSUE #584: endpoint for providing information about existing data to clients
User sessions may end up with autosave branches or existing PVs, depending on how they are terminated. We should introduce an endpoint that tells clients whether either of those is available for a given user / namespace / project / commit combination.

olevski:
closes #584 

Made according to the swagger spec that @lorenzo-cavazzi specified [here](https://app.swaggerhub.com/apis/lorenzo-cavazzi/renku-notebooks/0.1.0)

Example output:
```json
{
  "autosaves": [
    {
      "branch": "master",
      "commit": "993875e950de66a1074d58cb8ce89b54f41a0941",
      "date": "2021-03-08T16:04:08+00:00",
      "pvs": false
    }
  ],
  "pvsSupport": true
}
```

/deploy

### DIFF
--- a/renku_notebooks/__init__.py
+++ b/renku_notebooks/__init__.py
@@ -33,6 +33,7 @@ from .api.notebooks import (
     stop_server,
     server_options,
     server_logs,
+    autosave_info,
 )
 from .api.auth import bp as auth_bp, whoami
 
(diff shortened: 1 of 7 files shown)


rokroskar:
btw, you need `/deploy` not `\deploy` :)

lorenzo-cavazzi:
I haven't looked much into the code, but I've tested the endpoint in the PR deployment and it works well!
I've tried with projects in groups and subgroups, all good :tada: 

Would it be difficult to add also the support for PVS in the PR deplyoment, so that I can try also that part?

olevski:
@lorenzo-cavazzi I deployed this at https://tasko.dev.renku.ch/ with support for pvcs. So feel free to try that out.

lorenzo-cavazzi (approved):
Great, thanks! That works as advertised. And the code looks good too.

All fine from my side :heavy_check_mark:

PR #633, opened 2021-04-26
fix: add missing CHART_NAME var in ci image publishing step

olevski:
It seems this variable is required in several steps and is missing from one causing a new chart to not get published and fail.

See: https://github.com/SwissDataScienceCenter/renku-notebooks/runs/2418393909


olevski:
duplicate of #632

PR #637, opened 2021-04-27
fix: missing annotations handling in marshmallow

olevski:
This fixes a bug that resulted when an old session pod with some missing annotations was encountered. The bug happened in marshmallow and it resulted because annotations have `.` characters in their keys but marshmallow automatically converts keys like that in nested dictionaries. To bypass that the schemas that work with annotations had to be customized and some of the default marshmallow schema class methods had to be overridden. This overriding was not done properly and the overridden method was not taking into account missing/default keys and erroring out.

This fixes this problem.

I spun up a session in my deployment to test this. Having a pod with the annotations that are not marked as `required` in the marshmallow schema (such as `git-host`) does not cause the bug now. Also tested that having secrets without these annotations are not deleted but also do not crash the secret culling cronjob.

Currently deployed at https://tasko.dev.renku.ch/

### DIFF
--- a/renku_notebooks/api/schemas.py
+++ b/renku_notebooks/api/schemas.py
@@ -130,10 +130,10 @@ class UserPodAnnotations(
     class Meta:
         unknown = INCLUDE
 
-    def get_attribute(self, obj, key, *args, **kwargs):
+    def get_attribute(self, obj, key, default, *args, **kwargs):
         # in marshmallow, any schema key with a dot in it is converted to nested dictionaries
         # in marshmallow, this overrides that behaviour for dumping (serializing)
-        return obj[key]
+        return obj.get(key, default)
 
     @post_load
     def unnest_keys(self, data, **kwargs):


renku_notebooks/api/schemas.py
```diff
@@ -130,10 +130,10 @@ class UserPodAnnotations(
     class Meta:
         unknown = INCLUDE
 
-    def get_attribute(self, obj, key, *args, **kwargs):
+    def get_attribute(self, obj, key, default, *args, **kwargs):
```
rokroskar, line 133: where is `default` taken from?

olevski: @rokroskar default is passed by marshmallow depending on how the schema is specified. If the schema allows for missing values and it has defaults set then it will be used. I.e. https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.String if a default value is set then that value will be used.

olevski: Before what I had was just `obj[key]`. And this was erroring out because `git-host` was not marked as required in the schema and marshmallow expects you to have something that allows using a default value that will not error out (https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.Schema.get_attribute). The signature of the function in the docs specifically has the default value.

For a missing key that is not required marshmallow has some default default value that it uses and then it fully removes this field from the results. It makes sense since the field was not required and it has no default value it should not be shown at all in the results.

olevski: Also in the docs marshmallow has an example of overriding the `get_attribute` method in the same way as what I have in my fix here.

olevski: https://marshmallow.readthedocs.io/en/stable/extending.html#overriding-how-attributes-are-accessed

rokroskar: ok great, that makes sense to me

rokroskar (approved):

PR #641, opened 2021-05-03
fix: clusterrole is not required for user session PVCs

olevski:
The clusterrole that we added to the notebooks service for creating PVs for the user session is actually not required because the k8s storage plugin takes care of creating the underlying PV when the notebooks create a PVC. Therefore all that is needed for this functionality is a role and rolebinding that grant the permission to the notebooks service to manipulate PVCs.

Related to: https://github.com/SwissDataScienceCenter/renku/pull/2107


olevski:
I confirmed and tested this both with my personal dev deployment and with the acceptance tests on renku at https://github.com/SwissDataScienceCenter/renku/pull/2107.

olevski:
The helm chart that exludes this clusterrole is at https://swissdatasciencecenter.github.io/helm-charts/renku-notebooks-0.8.12-10b40b8.tgz

ableuler (approved):

Issue #648, opened 2021-05-10
enhancement: add sizeLimit to emptyDir used for user sessions


olevski:
This can be done like this in `spawners.py` at line 215:
```
if not options.get("pvc_name"):
    volume = {
        "name": git_volume_name,
        "emptyDir": {"sizeLimit": server_options.get("disk_request", "1G")},
    }
```

Pods are then evicted very quickly after the emptyDir goes over the limit:
```
  Normal   Created              6m20s  kubelet            Created container git-https-proxy
  Normal   Started              6m19s  kubelet            Started container git-https-proxy
  Warning  Evicted              62s    kubelet            Usage of EmptyDir volume "jupyter-tasko-2eolevskitestprivate-2dce351525-git-repo" exceeds the limit "1G".
  Normal   Killing              62s    kubelet            Stopping container notebook
  Normal   Killing              62s    kubelet            Stopping container git-https-proxy
```

However, the evicted status is not properly communicated to the UI s.t. the sessions can be deleted. Currently the UI shows the user sessions as if it was just stuck in "loading" and it does not provide the option to delete the session from the UI.

lorenzo-cavazzi:
If there is a way to get that information from the status we get from the `GET /servers`, we can easily adapt the UI to support that interaction. That works if the information is there some time (at least 1 min). Otherwise, it's a bit more complicated

olevski:
I will reopen a new issue to inform user when a session is evicted because of consuming too much memory, disk or other resources that have limits. Not sure how easy it is to do this since the eviction from k8s happen very quickly (i.e. within less than a minute for disk space).

PR #649, opened 2021-05-10
fix: properly assign default server option values

olevski:
This properly assigns the default values to server options in all cases.

The logic is as follows:
- You do not have to specify serverOptions at all in the api call to start a session. In this case all the defaults from the values.yaml are used. 
- If you do specify the server options, then the parameters for cpu, memory, default url and LFS auto-fetch are mandatory. The parameters for disk size and gpu are still allowed to be missing and replaced with defaults from values.yaml.
- The disk size request and gpus are not mandatory in values.yaml so in the case that the deployment does not include them in the values.yaml file then backup defaults are hardcoded in the renku notebooks code. These backup defaults are used only if the values.yaml file does not specify these parameters. The backup defaults are 0 gpus and 1G of disk space.

As for limiting the size when using emptyDir, I tested and k8s will evict the pods if they even go over the limit for their emptyDir by a few megabytes. When the limit is not respected the eviction happens in a matter of minutes if not seconds. However there are a few issues:
- The autosave branch is not created so all work is lost
- The status of the pod is `evicted` but the UI and notebooks do not really use or properly communicate this status. So a session is essentially stuck in limbo because it does not go away but its status is set to `evicted` but from the perspective of the notebooks it considers anything that is not with status `running` to be `loading`. The UI does not allow the user to shut down a session while it is `loading` so the `evicted` session can only be cleaned up by using `kubectl` or calling the API directly.

I have created a follow up issue to properly handle the size limit for emptyDir and cleanup of evicted sessions here #648.

One last question I have is what we want to do with the gpu and disk size portions for server options in values.yaml. As the code is written right now the cpu, memory, default url and LFS auto-fetch are mandatory in the values.yaml. Gpus and disk size are not mandatory and defaults for them are hardcoded in renku-notebooks code.

/deploy


ableuler:
Thanks for the immediate action on this! I have 4 questions/remarks:

1. 
> If you do specify the server options, then the parameters for cpu, memory, default url and LFS auto-fetch are mandatory. The parameters for disk size and gpu are still allowed to be missing and replaced with defaults from values.yaml.

Why not use the same logic for all parameters, ie allow all parameters to be missing and simply use the defaults from values.yaml for all missing params? I think that would be the simpler logic and would allow admins to for example disable showing the memory selection in the UI through the values file if they for some reason wanted to do this.

2. 
> These backup defaults are used only if the values.yaml file does not specify these parameters. The backup defaults are 0 gpus and 1G of disk space.

I would prefer to have these defaults (also) set in the values.yaml that we ship with the renku-notebook chart. The helm chart is currently the only way people deploy this application and the values.yaml is where people go first to look for configuration. If defaults are also set in the code, preferably only in the app config.

3. 
> The status of the pod is evicted but the UI and notebooks do not really use or properly communicate this status. So a session is essentially stuck in limbo because it does not go away but its status is set to evicted but from the perspective of the notebooks it considers anything that is not with status running to be loading. The UI does not allow the user to shut down a session while it is loading so the evicted session can only be cleaned up by using kubectl or calling the API directly.

I guess that is the case already now for pods that are evicted because of insufficient disk space on a node. So in that sense it's already an improvement over the status quo since this PR makes sure that the eviction hits a session that indeed consumes a lot of disk space.

4. One more thing to figure out for sure: do you know if the notebook containers usage of the empyDir volume is _also_ counted against the containers ephemearl-storage quota? I noticed that yesterday a pod without any limit on the empty dir got evicted when I added more than ~22GB of data to the emptyDir volume from within the session. This session had the default ephemeral-storage limit of 20GB. I'd find this a rather unexpected behaviour. But if what I describe above holds true, we might want to set the ephemeral storage limit on the notebooks container to 20GB + emptyDir limit to get a more expected behaviour.

olevski:
@ableuler I addressed all your comments from above. In addition to this I already implemented the changes required for #651 because it was easier to just include this rather than not and then add logic to parse and extract logic from the current way we provide the server options.

Now the values.yaml files have the following sections:
```yaml
serverOptionsUI:
  defaultUrl:
    order: 1
    displayName: Default Environment
    type: enum
    default: /lab
    options: [/lab]
  cpu_request:
    order: 2
    displayName: Number of CPUs
    type: enum
    default: 0.5
    options: [0.5, 1.0]
  # ....
serverOptionsDefaults:
  defaultUrl: /lab
  cpu_request: 0.5
  mem_request: 1G
  disk_request: 1G
  gpu_request: 0
  lfs_auto_fetch: false
```

The `serverOptionsUI` do not have to be mandatory, but it seems that the UI cannot handle server options that are fully empty because I think it always expects to have to render something. When I deployed with no server options for the UI the page for launching a new environment would never finish loading. I will submit an issue about this on the UI repo. So for now I made it so that at least the option for lfs auto fetch has to be provided in the values file - all other fields are optional.

Other than that all the other points we discussed are implemented:
- the defaults are always used and overwritten by any server options if any are passed in the post request for launching a session
- launching a session succeeds even if absolutely no server options are passed or even if an empty dictionary is passed for them (in this case the defaults from values.yaml are fully used)
- the ephemeral storage limit and disk size limit are both utilized - I added a fairly detailed explanation in the values file about this
- now the person who deploys renku can decide what server options to allow the user to change and which should not be changed and their only valid values are the values specified under `serverOptionsDefaults` in the values.yaml file.

I think this covers all use cases we may come across in the now and in the future.

olevski:
After this PR is merged then this PR should also be merged: https://github.com/SwissDataScienceCenter/renku/pull/2121

jupyterhub/spawners.py
```diff
@@ -212,7 +213,10 @@ def get_pod_manifest(self):
             volume for volume in self.volumes if volume["name"] != git_volume_name
         ]
         if not options.get("pvc_name"):
-            volume = {"name": git_volume_name, "emptyDir": {}}
+            volume = {
+                "name": git_volume_name,
+                "emptyDir": {"sizeLimit": server_options["disk_request"]},
+            }
```
ableuler, line 219: I think it would make sense to increase the pods limit and the request for ephemeral-storage by the emptyDir size limit here (if they are set). Then the value that is set through the charts would be the same independent of wether we're using a PVC or not (ie the amount of local storage that can be used for the the whole runtime environment).

helm-chart/renku-notebooks/templates/configmap.yaml
```diff
@@ -9,8 +9,10 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 data:
-  server_options.json: |
-    {{ toJson .Values.serverOptions }}
+  server_options_ui.json: |
```
ableuler, line 12: Could we also continue using `serverOptions` (instead of `serverOptionsUI`) for what's advertised to clients, and `serverDefaults` (instead of `serverOptions`) for the new defaults? This would avoid the breaking change.

ableuler (changes requested):

jupyterhub/spawners.py
olevski: I had to add some code to the spawner to achieve this. But it is done. I tested it - it works.

ableuler (approved):
Looks great, thanks for the updates! 👍

rokroskar:
@olevski any reason not to merge this?

olevski:
> @olevski any reason not to merge this?

No, no reason at all. I should have merged it earlier.

PR #657, opened 2021-05-14
fix: remove disk request values

olevski:
This removes the disk request values from showing up by default.

Also it addresses the separate issue of the notebook sessions not launching when the PVCs for user sessions are enabled but the disk request values are not shown in the UI.

#649 will properly address the multiple edge cases of what server options are used by default in which cases. This is just a quick fix to make sure the disk requests do not show up always and that if someone is using PVCs they will still be able to do things.


rokroskar (approved):

pameladelgado (approved):
Looks good :+1:

PR #668, opened 2021-05-28
K8s operator POC

olevski:
Depends on the following PRs:
- https://github.com/SwissDataScienceCenter/renku-gateway/pull/429
- https://github.com/SwissDataScienceCenter/renku/pull/2142

/deploy renku=notebooks-operator-poc


helm-chart/renku-notebooks/requirements.yaml
```diff
@@ -1,4 +0,0 @@
-dependencies:
-- name: jupyterhub
```
rokroskar, line 2: shouldn't this depend on amalthea?

helm-chart/renku-notebooks/values.yaml
```diff
@@ -118,158 +118,8 @@ serverDefaults:
   gpu_request: 0
   lfs_auto_fetch: false
 
-## Configuration for the jupyterhub service
```
rokroskar, line 121: add a stub for amalthea options here?

rokroskar (commented):
This is starting to look pretty good! A few minor comments in the code.

renku_notebooks/api/classes/server.py
```diff
@@ -161,219 +167,439 @@ def _get_image(self, image):
...
+        return output
+
+    def _get_session_k8s_resources(self):
+        cpu = float(self.server_options["cpu_request"])
+        mem = self.server_options["mem_request"]
+        gpu_req = self.server_options.get("gpu_request", {})
+        gpu = {"nvidia.com/gpu": str(gpu_req)} if gpu_req else None
+        resources = {
+            "requests": {"memory": mem, "cpu": cpu},
+            "limits": {"memory": mem, "cpu": cpu},
```
rokroskar, line 194: limiting the cpu is a departure from our current practice (we make it burstable usually by default). I think we should keep the current behavior for the time being but in a follow-up we could offer a setting (via the values file) to enforce cpu limits as well.

olevski: I think this is why starting the sessions is slow. My bad. I just completely overlooked the fact that the cpu should not be limited and only be part of the requests. I just copied and pasted the code for both requests and limits. I will change this - it was not a deliberate change.

renku_notebooks/api/classes/server.py
```diff
@@ -161,219 +166,439 @@ def _get_image(self, image):
...
+                    },
+                }
+            )
+            stateful_set_image_pull_secret_modifications.append(
+                {"name": image_pull_secret_name}
+            )
+        # Add git init / sidecar container
+        stateful_set_container_modifications.append(
+            {
+                "image": current_app.config["GIT_SIDECAR_IMAGE"],
```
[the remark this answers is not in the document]
olevski: Ok so this is def a problem. I added a sleep command in the bash script that clones the repository and then starts the git-sidecar. And I can definitely get into a session that has not cloned its repository. Even though the sleep command waits for the git server and and git proxy to become available it sleeps before starting to clone the repo. And the session does indeed become available to the user even though the repo has not been cloned. I will try to setup the readiness and liveness probes in k8s and see if this helps. Otherwise we will go with the solution that @rokroskar proposed here.

renku_notebooks/api/classes/server.py
olevski: done only memory is limited now

renku_notebooks/api/auth.py
```diff
@@ -42,92 +40,34 @@ def authenticated(f):
...
-                redirect(
-                    current_app.config["JUPYTERHUB_ADMIN_AUTH"].login_url
-                    + "&state=%s" % state,
-                    code=302,
-                )
-            )
-            response.set_cookie(
-                current_app.config["JUPYTERHUB_ADMIN_AUTH"].state_cookie_name, state
+            response = jsonify(
+                {"messages": {"error": "An authorization token is required."}}
```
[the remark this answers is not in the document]
olevski: Specified which headers need to be passed in for authentication

chartpress.yaml
```diff
@@ -7,36 +7,27 @@ charts:
       published: https://swissdatasciencecenter.github.io/helm-charts
     paths:
       - .
-      - jupyterhub
-      - singleuser
-      - cull_secrets
+      - renku-notebooks
```
ableuler, line 10: I think we want to add the git sidecar here too.

git-sidecar/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir --disable-pip-version-check -U pip && \
+    pip install --no-cache-dir --disable-pip-version-check pipenv
+
+COPY Pipfile Pipfile.lock clone.sh rpc-server.py /app/
+WORKDIR /app
+RUN pipenv install --system --deploy
+
+ENV HOST="0.0.0.0"
```
ableuler, line 17:
```suggestion
ENV HOST="127.0.0.1"
```

helm-chart/renku-notebooks/values.yaml
```diff
@@ -37,7 +44,7 @@ autoscaling:
 
 image:
   repository: renku/renku-notebooks
-  tag: 'latest'
+  tag: '0.8.13-9e28bc2'
```
ableuler, line 47: I guess that was not on purpose?

renku_notebooks/api/classes/server.py
```diff
@@ -157,224 +169,651 @@ def _get_image(self, image):
...
+                        "path": "/statefulset/spec/template/spec/containers/2/resources",
+                        "value": {
+                            "limits": {"memory": "64Mi"},
+                            "requests": {"memory": "32Mi"},
+                        },
+                    }
+                ],
+            }
+        )
+        if type(self._user) is RegisteredUser:
```
ableuler, line 617: i think this should go to amalthea

renku_notebooks/config.py
```diff
@@ -28,26 +27,9 @@
 IMAGE_REGISTRY = os.environ.get("IMAGE_REGISTRY", "")
 """The default image registry."""
 
-JUPYTERHUB_ANNOTATION_PREFIX = "hub.jupyter.org/"
+JUPYTER_ANNOTATION_PREFIX = "hub.jupyter.org/"
```
ableuler, line 30: Should we switch to an non-hub annotation?

ableuler (commented):
Had a first scroll through it and left some comments, will definitely need to spend a bit more time with it :)

renku_notebooks/api/classes/server.py
```diff
@@ -157,224 +169,651 @@ def _get_image(self, image):
...
+                "name": self.server_name,
+                "labels": labels,
+                "annotations": annotations,
+            },
+            "spec": {
+                "auth": session_auth,
+                "jupyterServer": {
+                    "defaultUrl": self.server_options["defaultUrl"],
+                    "image": self.verified_image,
+                    "rootDir": "/home/jovyan/work/",
```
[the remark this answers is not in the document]
ableuler: In principle we could solve this by querying the workdir directly from the image:
```
$ IMAGE=jupyter/base-notebook
$ TAG=latest
$ TOKEN=$(curl -s "https://auth.docker.io/token?scope=repository:$IMAGE:pull&service=registry.docker.io" | jq -r .token)
$ CONFIG_DIGEST=$(curl -s -H"Accept: application/vnd.docker.distribution.manifest.v2+json" -H"Authorization: Bearer $TOKEN" "https://registry-1.docker.io/v2/$IMAGE/manifests/$TAG" | jq -r .config.digest)
$ WORKINGDIR=$(curl -sL -H"Authorization: Bearer $TOKEN" "https://registry-1.docker.io/v2/$IMAGE/blobs/$CONFIG_DIGEST" | jq -r .config.WorkingDir)
$ echo $WORKINGDIR
 ```
So we could do this (slightly modified version for private images) to query the work directory from the image and then mount the volume at `$HOME/work`.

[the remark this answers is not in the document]
Comment (olevski) [reply renku_notebooks/api/classes/server.py]: Ok I added this functionality - we already query the docker api to see if the image truly exists and all that - don't see why we should not do this.

**Review by ableuler — APPROVED** [review]

**Comment by rokroskar** [comment]
if it's approved it's no longer a PoC 😆

File: helm-chart/renku-notebooks/values.yaml [hunk 12]
```diff
@@ -50,6 +57,23 @@ image:
...
 
+# oidc settings
+oidc:
+  clientId: renku-jupyterserver
+  clientSecret:
+  token_url:
+  auth_url:
+  allowUnverifiedEmail: false
+
+session_ingress:
```
ableuler, line 68:
```suggestion
sessionIngress:
```

ableuler: Noticed this when creating https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/96 that the notation here deviates a bit from the usual chart ingress notation that you get for example from `helm create`: 
 
 ```
 ingress:
  enabled: false
  annotations: {}
    # kubernetes.io/ingress.class: nginx
    # kubernetes.io/tls-acme: "true"
  hosts:
    - host: chart-example.local
      paths: []
  tls: []
  #  - secretName: chart-example-tls
  #    hosts:
  #      - chart-example.local
  ```

ableuler (changes requested):

helm-chart/renku-notebooks/values.yaml
olevski: 🤦  sorry yeah I will fix that. It seems I am somehow incapable of properly switching between snake, kebab and camel case.

olevski: @ableuler I found more cases in the `values.yaml` where I dont use camel case but I should. I will fix all of these with one PR.

ableuler: Using camelCase in yaml is definitely good, but my point was at leas as much about making hosts a list instead of single entry and expect the annotations as object rather than a string-serialized version of it.

ableuler: Actually, for the host vs hosts, I'd like to correct myself and keep it singular.

ableuler (approved):
🚀

PR #673, opened 2021-06-01
Add autosave delete endpoint

olevski:
This adds an endpoint for deleting an autosave branch or the associated PVC (if PVCs are used).

I also added a new class for an autosave which isolates some of the logic for the autosave out of the server class which is already getting quite overloaded IMO.

Lastly, I had trouble parsing the autosave and project names from the url path (i.e. both can have `/` in them). It seems that flask cannot distinguish very well when two path parameters can be side by side in the url. That is why I changed the endpoints a little bit so that the namespace/project comes first, then the `autosave` keyword and then the autosave name when you want to delete an autosave.

Let me know what you think. The logic for the naming of the autosave is as follows:
- if the autosave is a pvc, then the name of the autosave is the `{server_name}-pvc`, and this is also the name of the pvc in k8s
- if the autosave is a branch, then the name of this autosave is the branch name, the autosave branches already have a defined format for their names that we have been using
- in both cases querying the endpoint that lists the autosaves will also provide the correct autosave name which can be used to delete the branch or pvc

/deploy


ableuler (approved):
I like the factoring out of autosaves into their own class! 👍

lorenzo-cavazzi (approved):
I tried it in a few tricky scenarios and it worked very well! :+1: 

One minor thing I find a bit confusing on the autosave APIs (though it's not specifically for the DELETE endpoint): it's possible to end up in the situation below, where we have more than one autosave per branch/commit -- one is a branch, the other is a PVS.

Screenshot_20210604_165629

This should be rare (previous autosave and then a crash of the pod).
How does it behave in this case?
P.S. for the client it's not that relevant, it's just important to keep in mind there may still be multiple autosaves for the same branch/commit.

olevski:
@lorenzo-cavazzi what you show above is not unexpected. When you try to recover from this scenario the code does the following:
1. check if the right PVC exists, and if it does then recover the work from this PVC, delete the branch
2. check if the right autosave branch exists, and if it does exist then recover from it and delete it afterwards
3. if the expected PVC or autosave branch does not exist, then launch a new sessions without recovering any work
4. if a new sessions was launched successfully then go through any remaining PVCs and autosaves
  - if any of the remaining PVCs are tied to a parent commit of the commit that we are launching the session from AND they are not mounted to a session then delete them
  - if any of the autosave branches are tied to a parent commit of the commit that we are launching the session from then delete the branches

The "right" PVC or branch is based on the commit. In the example you posted if you launch a session from commit `88ddd8f` then all the steps start with 1 above are done. If you are launching from a commit that is a child of `88ddd8f` then steps 3 and 4 are done. And if you launch a session from a completely random and unrelated commit then 3 and 4 are done again but in this case step 4 will not delete any of the autosaves (pvc or branch).

PR #688, opened 2021-06-29
chore: tag notebook images with latest when tagging a release

olevski:
It seems that the notebook images are not tagged with latest at all. This should correct that. Now whenever images are tagged with a release tag the same images will also be published with the `latest` tag as well.

Here is the passing workflow for testing this: https://github.com/SwissDataScienceCenter/renku-notebooks/actions/runs/982073584

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,19 +73,12 @@ jobs:
         DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
     - name: Push latest image
       env:
-        CHART_NAME: renku-notebooks
         DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       run: |
-        if [ -z "$CHART_DIR" ]; then
-          CHART_DIR="helm-chart/"
-        fi
-        if [ ! -z "$IMAGE_PREFIX" ]; then
-          IMAGE_PREFIX="--image-prefix ${IMAGE_PREFIX}"
-        fi
         echo ${DOCKER_PASSWORD} | docker login -u ${DOCKER_USERNAME} --password-stdin
-        helm dep update $CHART_DIR/$CHART_NAME
-        chartpress --push --tag latest $IMAGE_PREFIX
+        helm dep update helm-chart/renku-notebooks
+        chartpress --push --tag latest
     - name: Wait for chart to get published
       run: sleep 120
     - name: Update component version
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -64,8 +64,8 @@ jobs:
       run: |
         echo "GIT_USER=Renku Bot" >> $GITHUB_ENV
         echo "GIT_EMAIL=renku@datascience.ch" >> $GITHUB_ENV
-    - name: Push chart and images
-      uses: SwissDataScienceCenter/renku/actions/publish-chart@0.7.11
+    - name: Push tagged chart and images
+      uses: SwissDataScienceCenter/renku/actions/publish-chart@master
```
ableuler, line 68: Hmm, I see why you have to do this, and it makes me realize that we should move this action out of the main Renku repo. We don't want to do a Renku release just for the action, but we don't want to use an untagged version here either, as this has broken things before. I'd propose a separate renku-actions repo. @rokroskar what do you think? Or do you have another idea as you have set many of those things up in the first place?

.github/workflows/ci.yml
```diff
@@ -74,7 +74,30 @@ jobs:
     - name: Wait for chart to get published
       run: sleep 120
     - name: Update component version
-      uses: SwissDataScienceCenter/renku/actions/update-component-version@0.7.11
+      uses: SwissDataScienceCenter/renku/actions/update-component-version@master
       env:
         CHART_NAME: renku-notebooks
         GITHUB_TOKEN: ${{ secrets.DEPLOY_TOKEN }}
+
+  publish-chart-latest:
```
ableuler, line 82: I'd rename this to "push-latest-image" as we're not publishing a "latest" chart, just pushing images.

.github/workflows/ci.yml
```diff
@@ -74,7 +74,30 @@ jobs:
     - name: Wait for chart to get published
       run: sleep 120
     - name: Update component version
-      uses: SwissDataScienceCenter/renku/actions/update-component-version@0.7.11
+      uses: SwissDataScienceCenter/renku/actions/update-component-version@master
```
ableuler, line 77: Same comment as above.

ableuler (changes requested):

.github/workflows/ci.yml
rokroskar: I was just reluctant at the time to make yet another repo - but yes, I agree it would make sense if those lived in a separate place.

ableuler: Ok, I've created this repo and tagged a first release that you can refer to in this PR: https://github.com/SwissDataScienceCenter/renku-actions/releases/tag/v0.1.0

olevski: That is great! Thanks @ableuler!

ableuler (approved):

PR #694, opened 2021-07-06
chore: install typed-ast so that black can work

olevski:
Updating black requires this package to be installed in python 3.7.

This will allow the tests for #682 to pass and the update to be applied.

I tried to change the dependabot PR itself but could not get past the conflicts in the Pipfile.lock. And dependabot says that if commits are added to the PR by someone else then the conflict resolution in the Pipfile.lock is turned off.

/deploy


ableuler (approved):
oh nice that you've figured that one out! 🎉

PR #700, opened 2021-07-16
fix: command and args for jupyterserver

olevski:
This is related to https://github.com/SwissDataScienceCenter/amalthea/pull/48.

It requires modifying the path because some of our images (i.e. r) has `tini` installed in `/` which is not in the path and others have it properly installed in `/usr/local/bin` or some other location which is in the path.

Currently deployed at tasko.dev.renku.ch.

Tested both with rstudio and jupyterlab type images.


olevski:
A single test is failing in the integration tests. It is about starting a session from different branches. It is something wrong with the test itself because trying this out on a deployment works without any issues. I will fix this.

olevski:
The tests were failing because the kind cluster would run out of resources and the sessions requires for subsequent tests could not be scheduled. This always occurred during the only test that needs to launch two sessions at the same time and check whether sessions from different branches can be launched. So the test was not the problem but the lack of resources. To fix this I lowered the default cpu/memory request that is used when launching a session and I split the tests so that they occur in different runners for the anonymous and registered sessions. Previously the tests for anonymous/registered sessions ran in the same runner at the same time - therefore making the problem with insufficient resources even worse.

PR #705, opened 2021-07-20
Dependabot Updates Batch

olevski:
The reason for the change in code is because of the update in the black package which caused some code formatting that passed before to fail now. Either way the changes are only in formatting.

/deploy


rokroskar (approved):
🤖

PR #708, opened 2021-07-22
fix: properly display storage

olevski:
This addresses a few bugs:
- Do not show ephemeral storage in the response that lists servers. The ephemeral storage is usually not related to what the user requested and can be set to a fixed value in the jupyterhub helm chart causing discrepancies between what the user asked when launching the session and what the api returns for the list of active servers.
- Use the username in the pvc. This was not the case before which would cause collisions between the pvcs for the same proejct/commit but for different users.
- Properly handle the data about disk size in the response when the api endpoint that list servers is called. This makes it so that things work with existing sessions from an old renku release where no disk size request has been set.

/deploy

### DIFF
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -100,3 +100,13 @@ def make_server_name(namespace, project, branch, commit_sha):
     return "{project}-{hash}".format(
         project=project[:54], hash=md5(server_string.encode()).hexdigest()[:8]
     )
+
+
+def make_pvc_name(safe_username, namespace, project, branch, commit_sha):
+    """Form a 16-digit hash server ID."""
+    server_string = f"{safe_username}{namespace}{project}{branch}{commit_sha}"
+    return "{username}-{project}-{hash}".format(
+        username=safe_username[:10],
+        project=project[:44],
+        hash=md5(server_string.encode()).hexdigest()[:8],
+    )
--- a/renku_notebooks/util/kubernetes_.py
+++ b/renku_notebooks/util/kubernetes_.py
@@ -103,7 +103,7 @@ def make_server_name(namespace, project, branch, commit_sha):
 
 
 def make_pvc_name(safe_username, namespace, project, branch, commit_sha):
-    """Form a 16-digit hash server ID."""
+    """Form a 16-digit hash persistent volume ID."""
     server_string = f"{safe_username}{namespace}{project}{branch}{commit_sha}"
     return "{username}-{project}-{hash}".format(
         username=safe_username[:10],
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,6 +48,7 @@ os.environ[
     "NOTEBOOKS_SERVER_OPTIONS_DEFAULTS_PATH"
 ] = "tests/dummy_server_defaults.json"
 os.environ["NOTEBOOKS_SERVER_OPTIONS_UI_PATH"] = "tests/dummy_server_options.json"
+os.environ["NOTEBOOKS_SESSION_PVS_ENABLED"] = "false"
 
 
 @pytest.fixture(scope="session", autouse=True)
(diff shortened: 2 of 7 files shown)


renku_notebooks/util/kubernetes_.py
```diff
@@ -100,3 +100,13 @@ def make_server_name(namespace, project, branch, commit_sha):
     return "{project}-{hash}".format(
         project=project[:54], hash=md5(server_string.encode()).hexdigest()[:8]
     )
+
+
+def make_pvc_name(safe_username, namespace, project, branch, commit_sha):
+    """Form a 16-digit hash server ID."""
```
rokroskar, line 106:
```suggestion
    """Form a 16-digit hash persistent volume ID."""
```

tests/conftest.py
```diff
@@ -337,7 +338,17 @@ def create_pod(username, server_name, payload):
                         }
                     },
                 }
-            ]
+            ],
+            "volumes": [
+                {
+                    "name": server_name + "-git-repo",
```
rokroskar, line 344: should this be using the pv name?

renku_notebooks/util/kubernetes_.py
```diff
@@ -100,3 +100,13 @@ def make_server_name(namespace, project, branch, commit_sha):
     return "{project}-{hash}".format(
         project=project[:54], hash=md5(server_string.encode()).hexdigest()[:8]
     )
+
+
+def make_pvc_name(safe_username, namespace, project, branch, commit_sha):
```
rokroskar, line 105: do we need two functions here for names? I think the reason `make_server_name` doesn't explicitly include the username is because the username is a part of the jupyterhub pod template - could we not change that so the template doesn't include the username, but then have just one function here that generates the unique string? We can then just attach `-pvc` to it to make it easier to identify.

rokroskar (commented):
Thanks @olevski this looks good - my main comment is about the code duplication with the new pvc name function - I'm wondering if we can refactor this a bit better by changing the jupyterhub templating.

tests/conftest.py
olevski: I would need to mock up the list of pvcs in k8s as well as adding/deletion when a session is started in order to make the units tests work with the PVCs. This was a lot easier. If you look at the top of the `conftest.py` file you will see that I specifically set the environment variable flag that indicates that PVCs are enabled to false. I am not sure it is worth pursuing this further because the tests change a lot with amalthea and a lot of these things that require extensive mocking move into the integration tests that use kind and an actual deployment.

renku_notebooks/util/kubernetes_.py
olevski: Yeah I know this sucks... But the server name is already being used in the annotation for each user session. So if I change the server name then I have to add a fair amount of code and if statements so that the server class is backwards compatible with the old server names that do not include the user name. 

I made a mistake in making the server name like that - it should have included the user name. So this was the easy way out so that I have to make the server class backwards compatible.

pameladelgado (approved):
should we merge this and make an issue to address @rokroskar 's comments at a later?
just to not stall the release, because not sure if addressing these would be easy/fast

olevski:
Yeah addressing the naming issue would require:
- an outage and getting rid of all existing user sessions that use an old naming scheme
- adding code to make the new release with the new naming scheme backwards compatible with the old naming scheme
- doing a "migration" of all existing user sessions where we directly modify all required parts of k8s resources - this would also still require an outage

I made a separate issue to address the naming here: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/709

I think it is best to tackle this in a separate PR and subsequent release.

rokroskar:
I don't see why it needs an outage? Doesn't the querying happen via pod metadata (which will remain unchanged)? In any case, I would expect this to be moot with the switch to Amalthea so I'm not sure it's worth the hassle right now.

PR #715, opened 2021-07-29
fix: do not fail on pvcs created with old insecure pvc naming

olevski:
With the bugfix from [renku 0.8.5](https://github.com/SwissDataScienceCenter/renku/releases/tag/0.8.5) I changed the way the PVCs for user sessions are named. However I did not fully account for "grandfathering" the PVCs that were created after the initial functionality in renku 0.8.4 was rolled out.

This caused a bug in the following scenario:
1. a user launched a session with renku 0.8.4 with a PVC
2. the 0.8.5 release was applied which changed how renku names PVCs
3. after the 0.8.5 release listing the servers from the notebooks api fails if there are any active sessions with PVCs created in 0.8.4

This failed only when trying to list the servers - specifically where the storage amount is added to the server response data here (see comments starting with `####`:
```
#### the session is using pvc but this evaluates to false because the server class uses the pvc name 
#### to check if the pvc exists or not
if server.session_pvc is not None and server.session_pvc.pvc is not None:       
exists - so the pvc is there but not under the right name so it is as if it does not exist
    resources["storage"] = server.session_pvc.pvc.spec.resources.requests[
        "storage"
    ]
else:
#### because the if statement above evaluated to false the code goes here where it expects an emptyDir 
#### but in reality there is a pvc and when it tries to access volume.empty_dir.size we get the error we have been seeing
    volume_name = pod.metadata.name[:54] + "-git-repo"
    volumes = [i for i in pod.spec.volumes if i.name == volume_name]
    if len(volumes) == 1 and volumes[0].empty_dir.size_limit is not None:
        resources["storage"] = volumes[0].empty_dir.size_limit
```

The fix here makes it so that if there is no emptyDir and the code ends in the else portion then there is no failure and information about storage is not added to the response.

I think this is the safest way to handle this rather than to test for the old-style PVC name and add the storage data. So for the naming of PVCs with the old method the API will not report storage at all. But I think this is fine considering there aren't that many of these anyhow. And as users launch new sessions they will have properly named PVCs and be able to see the storage they are using.

/deploy

### DIFF
--- a/renku_notebooks/api/schemas.py
+++ b/renku_notebooks/api/schemas.py
@@ -260,7 +260,11 @@ class LaunchNotebookResponse(Schema):
             else:
                 volume_name = pod.metadata.name[:54] + "-git-repo"
                 volumes = [i for i in pod.spec.volumes if i.name == volume_name]
-                if len(volumes) == 1 and volumes[0].empty_dir.size_limit is not None:
+                if (
+                    len(volumes) == 1
+                    and volumes[0].empty_dir is not None
+                    and volumes[0].empty_dir.size_limit is not None
+                ):
                     resources["storage"] = volumes[0].empty_dir.size_limit
             # remove ephemeral-storage if present
             if "ephemeral-storage" in resources.keys():


olevski:
p.s. The relevant part of the stack trace from this bug is as follows:
```
[2021-07-29 07:51:47 +0000] [8] [ERROR] Error handling request /jupyterhub/services/notebooks/servers
....
data = processor(data, many=many, **kwargs)
File "/renku-notebooks/renku_notebooks/api/schemas.py", line 282, in format_user_pod_data
"resources": get_server_resources(server),
File "/renku-notebooks/renku_notebooks/api/schemas.py", line 263, in get_server_resources
if len(volumes) == 1 and volumes[0].empty_dir.size_limit is not None:
AttributeError: 'NoneType' object has no attribute 'size_limit'
```

olevski:
... I am deploying renku 0.8.4, making a session, then deploying 0.8.5 and then this fix to confirm that I can replicate the bug and then to make sure it does not occur again

olevski:
I can replicate the bug. Testing if this new PR from notebooks will fix it. It only appears on project names that have a special character such as "-" in the name that `escapism` replaces with additional characters. If the project name is simple lowercase alphanumeric one then the bug actually does not appear because the pvc name is the same between renku 0.8.4 and 0.8.5 in this case.

olevski:
I can confirm this fix clears the bug and lets a user see and access all their sessions.

lorenzo-cavazzi (approved):
:+1:

PR #725, opened 2021-08-10
fix: improve messaging when cloning a repo runs out of hard drive space

olevski:
I am open to suggestions here, but I thought this may be a good idea.

Basically what we would do before is enable LFS autofetch immediately if the user requested and respect this setting essentially during the whole script that actually clones the repo. So when the session runs out of space and the cloning script fails then the user gets an empty repo without a message as to why things went wrong.

In order to quickly alleviate this without fully revamping our repo cloning script I have done the following:
- for as long as possible the repo cloning script avoids fetching LFS objects
- all the way at the end and if the user requested it the script will setup so that LFS objects are fetched by default and it will try to fetch the LFS objects
- if after fetching the LFS object the disk is 100% used the the local clone of the repo is deleted and a single file called `INSUFFICIENT_HARD_DRIVE_SPACE` is created in order to communicate the problem to the user

An alternative is to remove the part that removes the local copy of the repo and creates the file if the hard drive is full. This way if the hard drive fills up the user will get a (seemingly) full normal repo and a full hard drive. But some of the LFS objects will not all be there and as soon as the user tries to do something they will hopefully realize that they are out of hard drive space.

/deploy


olevski:
A few notes on alternatives (that are actually not viable):
- Using the information that gitlab reports at the top of a project that shows how much storage is used is not possible because this is the total of all storage files across all branches (I confirmed with a dummy project). So after the repository grows sufficiently this number is bound to become large and it is not a good representation for the real storage needs.
- Digging through the gitlab documentation on the API and the python SDK for gitlab I could not find anything about how to get the size of LFS objects for a specific branch or commit. There is only a way to get the total size of all LFS objects in the whole project just as I mentioned above, but this is not very useful.

I think to properly resolve this going forward we can use the git-sidecar and run a command in there that will list all LFS objects or simply check how much hard drive space is available.

Another alternative is that we revamp the bash script that clones the repository to use python. Then we can have the cloning script annotate the pod or jupyterserver resource in k8s with something that will indicate that there is insufficient hard drive space. The UI can then relay this information to the user.

rokroskar:
Would doing a checkout without checking out the LFS objects and then inspecting the pointer files in the tree work? The pointer files include the object size. You could parse the `.gitattributes` file to see which paths you need to check - I believe there is code already in renku-python that does this but maybe check with @Panaetius

olevski:
> Would doing a checkout without checking out the LFS objects and then inspecting the pointer files in the tree work? The pointer files include the object size. You could parse the `.gitattributes` file to see which paths you need to check - I believe there is code already in renku-python that does this but maybe check with @Panaetius

@rokroskar once I checkout the repo (even without fetch the LFS objects) I can get the size of all LFS files by doing `git lfs ls-files -s`. It would be nice if I did not have to do this to see how much size is required by the LFS objects.

The other problem is that once I know this in the init script that clones the repo, how do I communicate it to the notebook service. Or are you proposing that I should clone the repo without fetching the LFS objects in the notebook service itself when someone tries to start a session?

lorenzo-cavazzi:
> Another alternative is that we revamp the bash script that clones the repository to use python. Then we can have the cloning script annotate the pod or jupyterserver resource in k8s with something that will indicate that there is insufficient hard drive space. The UI can then relay this information to the user.

I can envision a very nice UX around having this information in the annotations (therefore in the `GET /notebooks` API) :slightly_smiling_face:

olevski:
>I can envision a very nice UX around having this information in the annotations 
It would be pretty easy to give the init container that clones the repo a simple k8s role that allows it to annotate the notebooks. So it would add an annotation to indicate that the session has run out of space.

Perhaps a better alternative is to open an endpoint on the notebook service that would result in annotating a user session. Then the git clone script could call that endpoint when it sees that the session runs out of space.

olevski:
Ok I added an issue to resolve this for the long term in #727. However for the time being we can provide some rudimentary info to users with this PR. All this does is that if the session runs out of space when the repo is being cloned then the whole repo is removed and a single filed called `INSUFFICIENT_HARD_DRIVE_SPACE` is created in the repository.

I think this is better than getting an empty repo with nothing in it which happens now when cloning the repo results in a full hard drive.

PR #729, opened 2021-08-12
feat: add s3 datasets

### LINKED ISSUE #627: s3 buckets minimal implementation

olevski:
Draft PR for the s3 dataset implementation depends on #668 

closes #627 

To whoever is looking at this - the biggest changes are represented by including a whole helm chart for datashim.io. Currently they do not have one so I took the templates they ship for installing things and converted them into a helm chart. I will reach out to datashim so that we can hopefully contribute this back to them and not have it in the notebooks repo.

But for now just to get going this is part of the notebooks repo.


rokroskar:
Could you make this PR on top of the operator branch instead of master so it's clearer what the new changes are?

olevski:
Currently this is deployed at https://s3poc.dev.renku.ch/.

See this slack post for more details: https://swiss-data-science.slack.com/archives/G67N59QL8/p1628941027012100

This deployment includes the following branches/features from other repos:
- https://github.com/SwissDataScienceCenter/renku-ui/tree/s3-datasets-poc
- https://github.com/SwissDataScienceCenter/renku-gateway/pull/429
- https://github.com/SwissDataScienceCenter/renku/pull/2142

olevski:
Deployed at https://s3poc.dev.renku.ch along with https://github.com/SwissDataScienceCenter/renku-ui/pull/1518

olevski:
@ableuler this is fully ready for review now. With the latest commit datasets can be turned on and off.

helm-chart/renku-notebooks/requirements.yaml
```diff
@@ -1,4 +1,8 @@
 dependencies:
 - name: amalthea
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: 0.2.2
+  version: "0.1.3"
+- name: dlf-chart
+  repository: "https://swissdatasciencecenter.github.io/datashim/"
+  version: "0.1.1-rc7"
```
ableuler, line 7: Feel free to publish an non-rc version of this chart before we merge this.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -10,11 +10,31 @@ global:
...
 amalthea:
   scope:
     clusterWide: false
   deployCrd: true
+  ## If enabling S3 datasets uncomment the section below
+  # extraChildResources:
+  #   - name: datasets
+  #     group: com.ie.ibm.hpsys
+
+## If enabling S3 datasets uncomment the section below
```
ableuler, line 25: I think these can be uncommented per default, if I'm not mistaken they will just be ignored if the datashim chart is not deployed.

renku_notebooks/api/classes/user.py
```diff
@@ -63,7 +73,8 @@ def __init__(self, headers):
         )
         if not self.authenticated:
             return
-        self.gitlab_client = Gitlab(current_app.config["GITLAB_URL"], api_version=4)
+        self.git_url = os.environ["GITLAB_URL"]
```
ableuler, line 76: afaik reading environment variables only in config.py and then use the app config in the application is considered best practice in flask.

renku_notebooks/api/classes/user.py
```diff
@@ -37,6 +39,14 @@ def jss(self):
         )
         return jss["items"]
 
+    @property
+    def datasets(self):
```
ableuler, line 43: Do we rely on the datasets being a user property? Would it be cleaner to treat them as a property of the user server?

renku_notebooks/api/classes/dataset.py
```diff
@@ -0,0 +1,181 @@
+from flask import current_app
+import boto3
+from botocore import UNSIGNED
+from botocore.client import Config
+from botocore.exceptions import EndpointConnectionError, ClientError, NoCredentialsError
+
+
+class Dataset:
```
ableuler, line 8: I understand that these are called datasets in the datashim project. However, at some point we want to distinguish them from our renku datasets. I wondering if this class could be the right point to make the "translation", by calling this class something like `S3mount` rather than dataset. wdyt?

helm-chart/renku-notebooks/values.yaml
```diff
@@ -10,11 +10,31 @@ global:
     domain:
   anonymousSessions:
     enabled: false
+  s3Datasets:
```
ableuler, line 13: I'd probably avoid the the dataset term here

ableuler (changes requested):
I think this is a well-crafted feature, looking forward to having it! I have a few minor comments and suggestions, nothing really fundamental.

helm-chart/renku-notebooks/values.yaml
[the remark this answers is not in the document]
olevski: @ableuler this should be a global value because when s3 dataset support is turned off the UI also needs to know about it so that it does not show the option to mount s3 buckets.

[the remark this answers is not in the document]
rokroskar: should there be an endpoint that lists the capabilities of the notebooks service instead?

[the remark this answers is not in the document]
olevski: > an endpoint that lists the capabilities of the notebooks service instead?
Yeah that is another alternative. @ableuler what do you think?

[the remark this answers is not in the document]
ableuler: Yes, I think that's the cleanest way. Conceptually it's very close to a server option. Should we use the same endpoint?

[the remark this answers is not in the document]
olevski: I will check with the UI team what they prefer. I think they directly turn every part of the server options into a UI element that is then used to select options by the user when they launch a session. This setting about the s3 mounting is more an overall setting for the whole delployment that cannot be turned no/off by the users. So because of that I think it may warrant putting into its own endpoint. Perhaps in an endpoint that also shows the current version of the notebook service and other high level info.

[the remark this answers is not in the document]
rokroskar: Yes that's what I had in mind like the version endpoint in the core service

helm-chart/renku-notebooks/requirements.yaml
olevski: I do not want to publish something like `0.1.1` because datashim.io do not publish their helm charts. I forked their repo in our org and added actions to publish the helm charts. So I am keeping the regular full release semvers to match regular releases from datashim.io. So I can publish something like `0.1.1-renku-1` or something along those lines. As long as it is a valid semver that helm likes. Helm can get quite picky about this.

ableuler: Do you expect datashim to publish a helm chart themselves anytime soon? Otherwise I I think publishing a datashim chart on the renku helm repo with say version `0.0.1` or `0.1.0` in which you refer to the latest datashim release as appVersion is acceptable. I'd expect chart and app version to diverge anyway, especially if we maintain the chart while they maintain the application.

renku_notebooks/api/classes/user.py
olevski: This is not really used. I will remove it. This is already a property of the user server.

helm-chart/renku-notebooks/requirements.yaml
olevski: >Do you expect datashim to take your changes upstream and publish a helm chart themselves anytime soon?

No this will not happen anytime soon. They do not publish their helm charts at all and are fully missing any CI pipeline to even publish their images.

Also the PRs that I have submitted to datashim change both the helm chart and the app. So I will figure out what is the best way to name the chart.

helm-chart/renku-notebooks/values.yaml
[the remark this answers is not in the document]
olevski: Now this is added to the `server_options` endpoint. Confirmed with the UI team they are ok with this.

helm-chart/renku-notebooks/requirements.yaml
olevski: Honestly I think the easiest thing to do is to just release a chart with tag with `rc` or something like what I just committed. Let me know what you think. The datashim helm charts are not the simplest. It is actually one parent helm chart with several children and some interesting templating. So I am not sure it is worth trying to figure all that out. I have already spent some time just to get the CI workflows that publish the helm charts at our fork over at https://github.com/SwissDataScienceCenter/datashim

olevski: Ok this was mostly due to a misunderstanding between me and Andreas on what datashim provides. The provide a whole helm chart but do not publish it or the images. We synced over slack and we are on the same page now.

ableuler (dismissed):

olevski:
@ableuler a minor edit was required to fix the integration tests. I forgot we have a bunch of test that verify the contents of the response from the `server_options` endpoint. And now we added a flag to indicate whether s3 mounts are enabled through the `server_options` endpoint.

ableuler (approved):

PR #742, opened 2021-09-06
fix: git proxy should let through unrecognized requests

olevski:
So after testing using git LFS on https://ci-renku-2142.dev.renku.ch/ I realized that things dont work.

And the reason is that the git proxy will block requests if the user is logged in and there are requests that are not directed to the git repo (i.e the git repos url). 

This essentially blocks some requests that git/git-lfs sends directly to the object storage that backs git LFS:
```
Prevented access to: https://os.zhdk.cloud.switch.ch/dev-lfs-objects/17/49/2d449dadb0d5cb10399b25c04fae8d44f62e1c511b90106689c4207e14ad?AWSAccessKeyId=XXXXXXXX&Signature=XXXXXXXXExpires=1630928506
```

This change should let through requests (without adding the auth headers for git) if the request is not for the git repository but the user is still logged in.

### DIFF
--- a/git-https-proxy/mitmproxy.js
+++ b/git-https-proxy/mitmproxy.js
@@ -64,25 +64,32 @@ proxy.onRequest(function (ctx, callback) {
    */
 
   // A bit annoying that we have to reverse-engineer the request url here...
-  let requestUrl =
+  const requestUrl =
     `${ctx.proxyToServerRequestOptions.agent.protocol}//` +
     ctx.clientToProxyRequest.headers.host +
     ctx.clientToProxyRequest.url;
 
 
+  const validGitRequest = ctx.proxyToServerRequestOptions.agent.protocol === repoUrl.protocol &&
+    ctx.proxyToServerRequestOptions.port === (repoUrl.port || defaultPort) &&
+    ctx.proxyToServerRequestOptions.host === repoUrl.host &&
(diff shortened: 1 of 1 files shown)


olevski:
Example project that fails: https://ci-renku-2142.dev.renku.ch/projects/tasko.olevski/large-lfs-dataset-2

git-https-proxy/mitmproxy.js
```diff
@@ -64,25 +64,32 @@ proxy.onRequest(function (ctx, callback) {
...
   if (
-    // User is not anonymous.
+    // User is not anonymous and request is for the git repository.
     // Important: make sure that we're not adding the users token to a commit
     // to another git host, repo, etc.
-    ctx.proxyToServerRequestOptions.agent.protocol === repoUrl.protocol &&
-    ctx.proxyToServerRequestOptions.port === (repoUrl.port || defaultPort) &&
-    ctx.proxyToServerRequestOptions.host === repoUrl.host &&
-    ctx.proxyToServerRequestOptions.path.startsWith(repoUrl.pathname)
+    !anonymousSession && validGitRequest
```
ableuler, line 92: We can only get here if both conditions are met, so you can get rid of the if clause.

ableuler (changes requested):
You can remove lines 100-104 as this section has become unreachable. The intent of this PR makes sense to me 👍

ableuler (approved):

PR #744, opened 2021-09-07
fix: add credentials in git proxy if host matches

olevski:
@ableuler I think we need to further relax the conditions under which we add the credentials in the git proxy.

As things currently stand (either in this branch or what we have in production) users cannot add another remote in the repository from the same git server. This is because another remote will have a different path (but the same host, port and protocol). 

This is probably not affecting a lot of users but if you want to pull changes from an upstream repository from which you have forked you have to do this. And when you try to do this (at least with an internal project) you have to provide a password and username for git. Even though the repository is on the same server.

Let me know what you think.

Here is an example internal project that I was testing with: https://renkulab.io/projects/tasko.olevski/maristem-omics

### DIFF
--- a/git-https-proxy/mitmproxy.js
+++ b/git-https-proxy/mitmproxy.js
@@ -72,8 +72,7 @@ proxy.onRequest(function (ctx, callback) {
 
   const validGitRequest = ctx.proxyToServerRequestOptions.agent.protocol === repoUrl.protocol &&
     ctx.proxyToServerRequestOptions.port === (repoUrl.port || defaultPort) &&
-    ctx.proxyToServerRequestOptions.host === repoUrl.host &&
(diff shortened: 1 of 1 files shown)


ableuler:
Restricting the proxy to work with only one repo is intentional. I know that this is annoying for legitimate use cases, but the security implications of removing this line would be too big. 
Just imagine evil Andreas knowing that Tasko has a repo `tasko/precious-project`. Then Andreas creates project `andreas/evil-project` with an image that will - say in its entrypoint - automatically try to push some malicious code to `tasko/precious-project`. Andreas then tricks Tasko into launching a session from `andreas/evil-project`, for example through some user support request, and without Tasko realising what's happening, Andreas has basically pushes malicious code changes to `tasko/precious-project`, to which he doesn't have access at all.

PR #750, opened 2021-09-15
fix: culling duration assignment

olevski:
This fixes a typo in the assignment of the culling idle threshold that made it so that all sessions (anonymous or registered) get assigned the same idle threshold of registered sessions.


olevski:
Will reopen this so that the list of commits is not crazy.

ableuler (approved):

PR #759, opened 2021-09-24
chore(tests): do not publish images for integration tests

olevski:
I think there were some issues with the kind action or I was not aware you could use chartpress to list all images and load them into kind and that is why I made the integration tests publish images. This should allow the integration tests to work without publishing images.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -84,21 +84,14 @@ jobs:
     - uses: actions/setup-python@v2
       with:
         python-version: 3.7
-    - name: Setup tmate session
-      uses: mxschmitt/action-tmate@v3
-      with:
-        limit-access-to-actor: true
     - name: Setup chartpress
-      env:
-        DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       run: |
         pip install --user pipx
         pipx ensurepath
         pipx install pipenv
         pipenv install --dev
-        echo ${DOCKER_PASSWORD} | docker login -u ${DOCKER_USERNAME} --password-stdin
-        pipenv run chartpress --force-build
+        pipenv run chartpress
+        kind load docker-image $(pipenv run chartpress --list-images) || true
         kind load docker-image $(pipenv run chartpress --list-images)
     - name: Install Helm Chart
       run: |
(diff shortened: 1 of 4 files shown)


ableuler (approved):

olevski:
@ableuler for some reason kind fails to load the images from the local docker the first time around. I tested this on my laptop and the same thing happened. But if I repeat the `kind load ...` command then it succeeds and using the image works without a problem.

So this is ready for review/approval now. I guess I should have started it out as a draft first.

olevski:
This will also fix the failing integration tests for the depandabot updates.

.github/workflows/integration-tests.yml
```diff
@@ -87,16 +85,14 @@ jobs:
...
-        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       run: |
         pip install --user pipx
         pipx ensurepath
         pipx install pipenv
         pipenv install --dev
-        echo ${DOCKER_PASSWORD} | docker login -u ${DOCKER_USERNAME} --password-stdin
-        pipenv run chartpress --push
+        pipenv run chartpress
+        kind load docker-image $(pipenv run chartpress --list-images) || true
```
ableuler, line 94: I think it would be good to open an issue on the kind repo (and maybe add a comment here that links to the issue).

ableuler (changes requested):
thanks for fixing this!

.github/workflows/integration-tests.yml
olevski: @ableuler I submitted an issue with kind here: https://github.com/kubernetes-sigs/kind/issues/2477

After trying to reproduce on my laptop I see that trying to load the images in kind all at once causes an error. Whereas trying to load the images withe separate commands, one by one, goes just fine.

.github/workflows/integration-tests.yml
```diff
@@ -87,16 +85,13 @@ jobs:
...
-        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       run: |
         pip install --user pipx
         pipx ensurepath
         pipx install pipenv
         pipenv install --dev
-        echo ${DOCKER_PASSWORD} | docker login -u ${DOCKER_USERNAME} --password-stdin
-        pipenv run chartpress --push
+        pipenv run chartpress
+        pipenv run chartpress --list-images | xargs -n 1 kind load docker-image
```
ableuler, line 94: 👍

ableuler (approved):

PR #771, opened 2021-10-05
fix(app): make size limit optional for empty dir user session storage

olevski:
By using the following field in the values.yaml file
```
userSessionPersistentVolumes:
  useEmptyDirSizeLimit: true
```

One can enable or disable imposing a limit on the empty dir (if empty dirs are used for user sessions).


pameladelgado (approved):
it looks good to me and was tested in a personal dev deployment
thanks! 👍

PR #794, opened 2021-11-05
fix: add ingress annotation to each session so iframes work

### LINKED ISSUE #793: Add nginx ingress annotation through a patch
Since we embed the sessions into a renkulab pages, it would make sense to add the following ingress annotation to the ingress objects which are created as part of the JupyterServer sessions: 
```
    nginx.ingress.kubernetes.io/configuration-snippet: |
        more_set_headers "Content-Security-Policy: frame-ancestors 'self' {{ .Values.ui.baseUrl }}"; 
```
Although we have already addressed this issue at the level of the session base images, adding it here would be more generic and independent of what's running inside the actual session.

olevski:
Fixes #793 

/deploy

### DIFF
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -733,13 +733,15 @@ class UserServer:
                 "patch": [
                     {
                         "op": "add",
-                        "path": "/ingress/metadata/annotations/-",
-                        "value": {
-                            "nginx.ingress.kubernetes.io/configuration-snippet": (
-                                'more_set_headers "Content-Security-Policy: '
-                                f"frame-ancestors 'self' {self.server_url}\";"
-                            )
-                        },
+                        # "~1" == "/" for rfc6902 json patches
+                        "path": (
+                            "/ingress/metadata/annotations/"
+                            "nginx.ingress.kubernetes.io~1configuration-snippet"
+                        ),
+                        "value": (
+                            'more_set_headers "Content-Security-Policy: '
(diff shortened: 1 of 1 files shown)


renku_notebooks/api/classes/server.py
```diff
@@ -727,6 +727,23 @@ def _get_session_manifest(self):
                     ],
                 }
             )
+        patches.append(
+            {
+                "type": "application/json-patch+json",
+                "patch": [
+                    {
+                        "op": "add",
+                        "path": "/ingress/metadata/annotations/-",
```
ableuler, line 736: Isn't that the syntax for adding to an array? The annotations are a map iirc.

olevski: Ah yeah that is true. My bad.

renku_notebooks/api/classes/server.py
```diff
@@ -727,6 +727,25 @@ def _get_session_manifest(self):
                     ],
                 }
             )
+        patches.append(
+            {
+                "type": "application/json-patch+json",
+                "patch": [
+                    {
+                        "op": "add",
+                        # "~1" == "/" for rfc6902 json patches
```
ableuler, line 736: ouff, good find!

ableuler (approved):

olevski:
p.s. The acceptance tests have been failing on launching anonymous sessions. But I tested this PRs deployment and I can launch anonymous sessions without a problem.

ableuler (approved):
LGTM 👍

PR #803, opened 2021-11-11
fix(chart): use current fallback renku image

olevski:
So yeah this slipped through the cracks a bit. I would have preferred to use a `latest` tab, but renkulab-py does not have `latest` tags.

### DIFF
--- a/helm-chart/renku-notebooks/values.yaml
+++ b/helm-chart/renku-notebooks/values.yaml
@@ -142,7 +142,7 @@ serverDefaults:
   lfs_auto_fetch: false
 
 ## Default image used when the requested image for a user session cannot be found.
-defaultSessionImage: "renku/singleuser:latest"
+defaultSessionImage: "renku/renkulab-py:3.9-0.10.1"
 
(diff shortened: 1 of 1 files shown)


rokroskar:
We can make it have the latest tag, but then you run the risk of running an image that's actually not compatible with the instance you're on.

rokroskar:
Maybe we can add some automation in the renkulab-docker repo to automatically update this whenever a new tag is made there.

olevski:
> We can make it have the latest tag, but then you run the risk of running an image that's actually not compatible with the instance you're on.

Yeah that is a good point... I just have to remember to bump the fallback image in the helm chart whenever we update the images.

rokroskar:
You could do that every time or spend ~an hour~ a day implementing https://github.com/SwissDataScienceCenter/contributed-project-templates 😆

ableuler (approved):

ableuler:
My ideal scenario would be to not maintain any base-images anymore, just templates.

Issue #800, opened 2021-11-10
Add node affinity / node selector to user pods


ableuler:
Currently, user pods only carry a toleration for the default renku user node taint, but no corresponding node affinity or node-selector is added. Hence, when user nodes fill up, user sessions are happily scheduled onto non-user nodes. Ideally, it would be fully configurable through the chart along those lines:
- which labels must be present on a node for a user session to be allowed to run on it (default none)
and
- which taints can be present on a node for a user-session to be still run on it (ie tolerate it) (default: `renku.io/dedicated=user:NoSchedule`)

ableuler:
closed by #806

PR #806, opened 2021-11-15
feat(chart): add session tolerations, affinity, nodeSelector

### LINKED ISSUE #799: Tolerations and node affinity for sessions should be configurable
Currently we only assign tolerations and they are essentially hardcoded:

```
tolerations = [
            {
                "key": f"{current_app.config['RENKU_ANNOTATION_PREFIX']}dedicated",
                "operator": "Equal",
                "value": "user",
                "effect": "NoSchedule",
            }
        ]
```

And we do not do anything about affinities. We should assign both and make this an option to adjust/change in the values file.

olevski:
This adds the ability to specify tolerations, affinity, etc specifically for the user sessions.

I tested with specifying:
- node selector and tolerations
- affinity and tolerations

Note that specifying an affinity for one set of nodes and then node selector for another resulted in an unschedulable pod.

Deployed at tasko.dev.renku.ch and set up with affinity and tolerations for the user dedicated nodes.

As implemented the tolerations, affinities and nodeSelector fields that are by default currently empty as they come from Amalthea are fully replaced - not merged. So if at some point Amalthea assigned any of these fields for a user sessions they would be replaced in the notebook service. I think this is a safe assumption though - we really try to avoid getting in the way with Amalthea and letting the users do their thing. So I don't believe we would ever go for imposing any of these things on the session level in Amalthea.

Lastly by default I left the values file blank. I think it is easiest that way - we do not know for sure what all renku deployments have setup. This means that all existing deployments should update this field in their values file accordingly and as required.

closes #799 #800


ableuler (approved):
LGTM! 👍

pameladelgado:
> Lastly by default I left the values file blank.

I see that the code is taking an empty value as default, I guess those would be the default values, right?

pameladelgado:
we could include some commented lines about how to configure notebooks to enable session-dedicated nodes, in the values file here and in renku's values file.

ableuler:
> maybe I would consider including in the variables names `TAINTS_TOLERATIONS` and `NODE_AFFINITY`, it's a bit more verbose but clears any doubt on what it might be referring to.

@aledegano here I would actually prefer to stay with the naming of the fields in the pod spec that we're effectively setting. Especially for the affinity there is room for confusion since `nodeAffinity` is a sub-field of `affinity`. What we're setting here is the affinity and in principle there's nothing stopping a user from defining eg an anti-affinity to another pod. Therefore I think adding the option to modify the entire `affinity` block and not just a `nodeAffinity` is the right way to go.

olevski:
> > Lastly by default I left the values file blank.
> 
> I see that the code is taking an empty value as default, I guess those would be the default values, right?

@pameladelgado by default the values file sets empty lists of dictionaries. I said blank in the description but that is what I meant.

olevski:
> we could include some commented lines about how to configure notebooks to enable session-dedicated nodes, in the values file here and in renku's values file.

I am sorry I rushed the merge. I will add a bit more comments in the values file in the PR for the new renku notebooks release.

PR #814, opened 2021-11-18
fix(app): do not delete autosaves before restoring

olevski:
I wish you could hide PRs...

### DIFF
--- a/renku_notebooks/api/classes/storage.py
+++ b/renku_notebooks/api/classes/storage.py
@@ -27,7 +27,11 @@ class Autosave:
                 f"projects/{self.gl_project.id}/repository/merge_base",
                 params={"refs[]": [self.root_commit_sha, commit_sha]},
             )
-            if res.status_code == 200 and res.json().get("id") == self.root_commit_sha:
+            if (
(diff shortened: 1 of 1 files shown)


ableuler (approved):
This fixes the bug. Let's 🚢 -it

ableuler:
> I wish you could hide PRs...

Haha, you didn't exactly go stealth in the naming of the PR. `fix: restore expected autosave behaviour` would have been the other option :)

olevski:
I tested the following scenarios:
- launch a session, add files, shut it down, recover work (session_commit=HEAD, autosave_root_commit=HEAD)
- same as the above except add files and commit them (but do not push the changes)
- same as the above but with a combo of committed and uncommitted work
- launch a session, add stuff, commit and push, then add more file, then shut session down, then launch a session from commit HEAD-1 (session_commit=HEAD-1, autosave_root_commit=HEAD-1) - in this case the work is restored
- same as above except launch a session from HEAD (session_commit=HEAD, autosave_root_commit=HEAD-1) in this case the work is not recovered and the autosave branch is deleted

PR #819, opened 2021-11-22
chore: fix ci deployment pipeline

ableuler:
/deploy


olevski (approved):

olevski:
The failing acceptance tests are an unrelated issue btw. This change does not affect that. The deployment is successful that is all that matters with this change. So we should just merge this even with the failing acceptance tests.

PR #823, opened 2021-11-25
chore: update renku actions to 0.3.1

olevski:
/deploy


Panaetius (approved):
Thank you!

olevski:
@Panaetius thanks for the approval. But we recently found a bug in 0.3.1 in the action that deploys renku. Turned out to be because of non-posix compliant if statement. Anyhow I will close this and re-open a PR for 0.3.2. For now the action to deploy renku in 0.3.0 works.

PR #828, opened 2021-11-30
fix(app): use 3 scenarios for session cpu limits

olevski:
As discussed. I deployed this 4 different times for each case and tested that it works.


ableuler (approved):
LGTM! 👍

PR #834, opened 2021-12-02
fix(app): return proper values for image and repository

### LINKED ISSUE #810: Return the repository URL in the annotations
This was removed at some point but it is used by the UI.

This is how the UI uses this value:
`${annotations["repository"]}/tree/${annotations["branch"]}`

olevski:
/deploy

closes #810

### DIFF
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -314,6 +314,9 @@ class UserServer:
             current_app.config["RENKU_ANNOTATION_PREFIX"]
             + "projectName": self.gl_project.path.lower(),
             f"{current_app.config['RENKU_ANNOTATION_PREFIX']}requested-image": self.image,
+            f"{current_app.config['RENKU_ANNOTATION_PREFIX']}repository": None
+            if self.gl_project is None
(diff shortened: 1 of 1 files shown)


renku_notebooks/api/classes/server.py
```diff
@@ -314,6 +314,9 @@ def _get_session_manifest(self):
             current_app.config["RENKU_ANNOTATION_PREFIX"]
             + "projectName": self.gl_project.path.lower(),
             f"{current_app.config['RENKU_ANNOTATION_PREFIX']}requested-image": self.image,
+            f"{current_app.config['RENKU_ANNOTATION_PREFIX']}repository": None
+            if self.gl_project is None
```
Panaetius, line 318: There's a couple of places above this line, e.g. `current_app.config["RENKU_ANNOTATION_PREFIX"] + "projectName": self.gl_project.path.lower()` where we also access properties of `gl_project` which could be None, in which case the code would fail. I think this should be handled on those lines, too.

Panaetius (changes requested):

Panaetius (approved):

PR #841, opened 2021-12-02
chore(app): Update default session image to renku/renkulab-py:3.9-tast-auto-update-10

RenkuBot:
Generated automatically on a new release in renkulab-docker to update the default image used when launching a user session.


olevski:
DO NOT MERGE - just a test

PR #844, opened 2021-12-06
fix(app): disable renku cli version check

olevski:
Prevents the renku cli from checking for the latest version and printing a message about the upgrading every time a user runs a renku cli command.


Panaetius (approved):
Thank you!

olevski:
@Panaetius I am sorry I should have closed this - I kind of forgot about it and left it open. Rstudio ignores the env variables set in the container. The environment variables for Rstudio have to be added in the docker image. I am testing changes there and will submit a PR in that repo.

PR #845, opened 2021-12-07
fix(app): do not accept an empty anonymous user id

olevski:
The gateway defaults to anonymous user id set to `""` when one is not found in the cookie for a users session. But having a token set to `""` results in improper k8s resource names. I.e. the sessions are named like `f"{user-id}-some-other-snippet"`. And when the anonymous user id is blank the resources in k8s are named starting with `-` and k8s does not allow this.

Also because of k8s naming restrictions and how the id is used in naming resources the anonymous id must start with an alphanumeric character. I added a check for this too.

/deploy (just to be sure)


ableuler (approved):
LGTM!

PR #848, opened 2021-12-13
feat(app): enable age-based session culling

olevski:
This is required so that we enable age-based culling for anonymous sessions in different deployments.

By default I made it so that age-based culling is turned off for anonymous and registered sessions. This is more in line with past behaviour.


pameladelgado (approved):
LGTM!

PR #849, opened 2021-12-13
enhancement(app): standardized error responses

### LINKED ISSUE #545: enhancement: standardized error responses
All errors from the API should adhere to the following schema:

```
class ErrorNested(Schema):
    message: fields.Str(required=True)
    detail: fields.Str(default: "")

class ErrorResponse(Schema):
    error = fields.Nested(
        ErrorNested(), required=True
    )
```

With this an error would look like this:
```
{
"error": {
       "message": "Invalid arguments.",
       "detail": "Argument A should be an integer.",  // this field can also be just blank ""
    }
}
```

A few questions:
- You mentioned you would also include a `code` field in the `ErrorNested` part. How do I come up with this code and how would you use it in the UI? Do we need another agreement between the UI and backend about what these codes mean?
- What happens if there are multiple errors? Should the `error` field be actually a list of the errors that were triggered? In that case should the error field always be a list?
- Is it ok to not have a detailed message in all cases i.e. `default=""`?

olevski:
/deploy #persist renku-ui=master

closes #545 

This will allow us to address some more recent issues about communicating to users better what went wrong when trying to launch a session and things fail.


renku_notebooks/api/schemas.py
```diff
@@ -553,3 +544,23 @@ class AutosavesList(Schema):
 
     pvsSupport = fields.Bool(required=True)
     autosaves = fields.List(fields.Nested(AutosavesItem), missing=[])
+
+
+class ErrorResponseNested(Schema):
+    """Generic error response that can be used and displayed by the UI.
```
ableuler, line 550: What's the reasoning behind introducing this mapping instead of just using the matching http error codes?

renku_notebooks/api/notebooks.py
```diff
@@ -261,26 +287,38 @@ def autosave_info(user, namespace_project):
...
             404,
         )
     autosave = Autosave.from_name(user, namespace_project, autosave_name)
     if not autosave.exists:
         return make_response(
             jsonify(
-                {"messages": {"error": f"The autosave {autosave_name} does not exist"}}
+                {
+                    "error": {
+                        "message": f"The autosave {autosave_name} does not exist",
```
ableuler, line 318:
```suggestion
                        "message": f"The autosave branch {autosave_name} does not exist",
```
Now we're talking only about branches, right?

renku_notebooks/api/notebooks.py
```diff
@@ -261,26 +287,38 @@ def autosave_info(user, namespace_project):
...
         204: {"description": "The autosave branch and/or PV has been deleted."},
-        404: {
-            "description": "The requested project, namespace and/or autosave cannot be found."
-        },
     },
 )
+@marshal_with(
+    ErrorResponse(),
+    code=404,
+    description="The requested project, namespace and/or autosave cannot be found.",
```
ableuler, line 295:
```suggestion
    description="The requested project, namespace and/or autosave branch cannot be found.",
```

ableuler (changes requested):
Definitely much nicer error handling! 

Not sure if I'm missing the use case, but in the present form I do not see much value in using the custom error codes on top of the HTTP status codes. If we introduce our own error codes, I think they should uniquely determine one and only one error. This would for example mean that we use a different error code for a generally invalid request header/parameter than for the case where specifically the username is too long. 
At the moment I find the message that you include in the error response much more specific/useful than what I'd get through a lookup here:
```
    Status code for errors are as follows:
    - 10: failed parsing of request data/params - maps roughly to 422
    - 20: the requested resource or a related resource is missing - maps roughly to 404
    - 30: the required k8s operation to be exucted failed - maps roughly to 500
    - 40: authentication error - maps roughly to 401
```
so I'm not sure what the use case of those error codes is.

renku_notebooks/api/schemas.py
olevski: In the design of the error response schema the UI wanted to have some kind of a code. So I added this.

lorenzo-cavazzi:
That code is unrelated to the HTTP status code. Those are generally 401, 404 and 50* and don't really provide much information about how to address the underlying error.

The idea of the error code is to easily identify the same type of error. E.G. a wrong `git` operation, even if it happens on different APIs and within a different context.
This seems an effective way of dealing with low-level errors where messages are not user-friendly (this is where we were heading with this on renku-core SwissDataScienceCenter/renku-python#2303).

If you think it's not useful (or feasible) for renku-notebooks, it's fine to leave it out. It may need a few more customize user-friendly error messages at the API level to give meaningful information to users since the UI won't be able to replace them.

olevski:
@lorenzo-cavazzi I am just curious, how does the UI use these error codes? I guess in some cases the styling of the message is different? Or some error codes are not shown to the user - because there is nothing the user can do about that specific type of error?

olevski:
I think we have a few ways to move forward:
1. remove the codes and probably add them in a subsequent PR
2. expand this PR to give more detailed codes (similar to what Lorenzo pointed out in the PR)
3. keep the current codes - and only display codes 10, 20, 40 to the user - with 30 resulting in some generic message that something is wrong or something like that

I think that option 1 makes most sense from a practical point of view.

olevski:
One thing that I should (and will) do right now is to refactor things a little bit so that the error handling is centralized in one handler just like how it is in renku-python.

olevski:
Ok I added similar type of error codes and handling like what Lorenzo pointed out in the link to the renku python PR he sent. This is now ready as far as I am concerned. It could be that the error codes require a bit more polishing but I think that this is a good start. I reused the error codes and general types of errors from PR Lorenzo pointed out.

ableuler:
The sibling PR of this one is https://github.com/SwissDataScienceCenter/renku-python/pull/2303, right?

renku_notebooks/api/classes/server.py
```diff
@@ -70,13 +71,13 @@ def __init__(
     def _check_flask_config(self):
         """Check the app config and ensure minimum required parameters are present."""
         if current_app.config.get("GITLAB_URL", None) is None:
-            raise ValueError(
-                "The gitlab URL is missing, it must be provided in "
+            raise ProgrammingError(
```
Panaetius, line 74: Should this be its own error type, like `ConfigurationError` ?

Because the description of 
> The programming errors are bugs or unexpected cases. In the first case, they should lead
    to creating a new GitHub issue; in the latter, it may be necessary to handle the specific
    error to provide the user a precise explanation.

Doesn't really match this issue

renku_notebooks/api/classes/server.py
```diff
@@ -70,13 +71,13 @@ def __init__(
...
+                message="The gitlab URL is missing, it must be provided in "
                 "an environment variable called GITLAB_URL"
             )
         if current_app.config.get("IMAGE_REGISTRY", None) is None:
-            raise ValueError(
-                "The url to the docker image registry is missing, it must be provided in "
+            raise ProgrammingError(
```
Panaetius, line 79: Same as above

renku_notebooks/api/schemas.py
```diff
@@ -592,3 +584,93 @@ class AutosavesList(Schema):
...
+
+class ErrorResponseFromGenericError(ErrorResponse):
+    @pre_dump
+    def extract_fields(self, err, *args, **kwargs):
+        response = {
+            "message": err.message,
+            "code": err.code,
+        }
+
+        if hasattr(err, "detail") and err.detail is not None:
```
Panaetius, line 617:
```suggestion
        if getattr(err, "detail", None) is not None:
```

renku_notebooks/api/schemas.py
```diff
@@ -592,3 +584,93 @@ class AutosavesList(Schema):
...
+        try:
+            code = self.status_code_map[err.code]
+        except KeyError:
+            pass
+        response = {
+            "message": err.description,
+            "code": code,
+        }
+
+        if hasattr(err, "detail") and err.detail is not None:
```
Panaetius, line 669:
```suggestion
        if getattr(err, "detail", None) is not None:
```

renku_notebooks/api/schemas.py
```diff
@@ -592,3 +584,93 @@ class AutosavesList(Schema):
...
+        505: 1505,
+    }
+
+    @pre_dump
+    def extract_fields(self, err, *args, **kwargs):
+        code = 2500
+        try:
+            code = self.status_code_map[err.code]
+        except KeyError:
+            pass
```
Panaetius, line 663:
```suggestion
        code = self.status_code_map.get(err.code, 2500)
```

Panaetius (dismissed):

olevski:
For anyone coming back to this the schema for all error responses from the notebook service is:

```python
class ErrorResponseNested(Schema):
    """Generic error response that can be used and displayed by the UI."""

    message = fields.Str(required=True)
    detail = fields.Str(dump_default=None)
    code = fields.Integer(required=True)
```

The `message` and `code` fields are always present, `detail` can be null.

lorenzo-cavazzi (changes requested):
Thanks for putting this together, and sorry for letting you wait so long before implementing the client-side!

Compared to the renku-python solution, here the messages seem to be always targeting the user, right? I kinda liked the `userMessage` vs `devMessage` distinction, but maybe it's not necessary after all.
Would it make sense to include the sentry logic in the error handling so that you can return the sentry exception link in the payload? That might be useful for devs. Something like this https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/errors.py#L87-L133
Even though I would leave that for a follow-up issue so we don't keep this pending.

The code already looks great! I added just a few inline comments, nothing critical.

renku_notebooks/errors/utils.py
```diff
@@ -0,0 +1,25 @@
...
+        response = e.get_response()
+        # replace the body with JSON
+        response.data = ErrorResponseFromWerkzeug().dumps(e)
+        response.content_type = "application/json"
+        return response
+    elif isinstance(e, GenericError):
+        return ErrorResponseFromGenericError().dump(e), e.status_code
+    else:
+        # now you're handling non-HTTP exceptions only
+        # TODO: Log
```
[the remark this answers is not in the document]
olevski: That is for a future issue. But I am not sure if we even need it to begin with. So for now I would like to leave it as is.

renku_notebooks/errors/user.py
```diff
@@ -0,0 +1,51 @@
...
+    status_code: str = 422
+
+
+@dataclass
+class MissingResourceError(UserInputError):
+    message: str = (
+        "The resource you are requesting or is needed for your request does not exist."
+    )
+    code: int = 1404
+    status_code: int = 404
```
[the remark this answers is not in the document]
olevski: Yeah should be an int. Will change it.

renku_notebooks/errors/user.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    code: int = 1401
+    status_code: int = 401
+
+
+@dataclass
+class DuplicateS3BucketNamesError(UserInputError):
+    message: str = "The names of all mounted S3 buckets should be unique."
+    code: int = 1402
+    status_code: int = 422
```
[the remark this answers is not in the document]
olevski: My bad will remove that. No reason to repeat things.

renku_notebooks/errors/intermittent.py
```diff
@@ -0,0 +1,43 @@
...
+    ] = "If this problem persists please contact your administrator."
+
+
+@dataclass
+class DeleteServerError(IntermittentError):
+    message: str = (
+        "The server cannot be deleted, most likely due to problems "
+        "with the underlying infrastructure."
+    )
+    code: int = 3001
```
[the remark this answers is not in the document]
olevski: Sure. I will do that across the board.

renku_notebooks/errors/user.py
```diff
@@ -0,0 +1,51 @@
...
+
+    message: str = "Invalid user input."
+    code: int = 1000
+    status_code: str = 422
+
+
+@dataclass
+class MissingResourceError(UserInputError):
+    message: str = (
+        "The resource you are requesting or is needed for your request does not exist."
```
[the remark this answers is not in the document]
olevski: I will remove the default. I agree.

olevski:
@lorenzo-cavazzi as for including a sentry issue url I created a follow up issue.

I think this is really not critical. I have no issues distinguishing or tracking down specific issues in Sentry. I also do not want to slow down the API response time by having to call Sentry before I send the response to the user. But that is a separate discussion topic.

lorenzo-cavazzi (dismissed):
Lgtm! Thanks for the effort :rocket:

lorenzo-cavazzi (approved):

lorenzo-cavazzi (approved):

PR #858, opened 2021-12-18
fix(app): case insensitive project and namespace

olevski:
Addresses problems arising from changing the path of a project. Or from confusion between case sensitive differences where they do not matter (based on gitlab).


olevski:
Addresses an issue a user came across:
https://gitter.im/SwissDataScienceCenter/renku?at=61bc5b8b0c681346aeedb20f

resource_migrations/1.3.0.py
```diff
@@ -0,0 +1,111 @@
+import argparse
```
ableuler, line 1: It's very nice that you provide this migration option. There is a strong similarity to the case of an application using a DB. This makes me wonder: do you just want to call these migrations in either the entrypoint of the main container or a separate init container of the notebook-service?

olevski: @ableuler I think an init container would make a lot of sense tbh. I will make that change.

olevski: That way I also don't have to update the docs here and in renku on how to run such migrations. Also I can keep adding such migrations to this init container.

olevski: Done @ableuler. I tested on my dev deployment a case where no patching is needed and a case where the project name needs to be patched. It worked.

resource_migrations/migration_1.py
```diff
@@ -0,0 +1,67 @@
...
+                        name=js_name,
+                        body={
+                            "metadata": {
+                                "annotations": {annotation_key: annotation_val.lower()}
+                            }
+                        },
+                    )
+            else:
+                print(f"No need to patch {js_name} for annotation {annotation_key}")
+            sleep(2)
```
ableuler, line 62: I think this is problematic. If I parsed the indents correctly, this will lead to a runtime of 2s * number_of_migrations * number_of_sessions * number_of_annotation_keys of the init container, even when all jupyter servers are already up-to-date.

If we add a sleep to protect the API from being overloaded, it should only take place after a patch request. But I think the API call in `k8s_api.patch_namespaced_custom_object` is synchronous anyway and will thus block the code execution if the API is overloaded, so I'd go without any sleep.

resource_migrations/migration_1.py
```diff
@@ -0,0 +1,67 @@
...
+from kubernetes import config as k8s_config
+
+from run_all import parse_args
+
+
+def adjust_annotations(args):
+    k8s_config.load_config()
+    k8s_api = client.CustomObjectsApi(client.ApiClient())
+
+    jss = k8s_api.list_namespaced_custom_object(
```
ableuler, line 13: The API supports pagination, so I think it would be good to use it (probably with a rather large page size).

resource_migrations/run_all.py
```diff
@@ -0,0 +1,67 @@
...
+        help="The renku k8s annotation prefix.",
+    )
+    args = parser.parse_args()
+    return args
+
+
+def run_all(args):
+    # Run all migrations in order
+    print("Starting k8s resource migrations.")
+    migration_1.adjust_annotations(args)
```
ableuler, line 62: How about adding the "migration version" as a label to each notebook server? Then the `k8s_api.list_namespaced_custom_object` query could filter by this label and thus only select the objects which still need to apply the current migration.

ableuler (changes requested):
I like the strategy "upgrades are handled automatically, for downgrades kill all sessions". I think that's very reasonable. 
I have some concerns about the practical application of the migrations though. There's the potentially long runtime due to the sleep statements (which I think can easily be fixed), and I'm wondering what happens if multiple pods try to execute a migration simultaneously I guess it would probably still work as multiple pods would just try to apply the same PATCH request to the same resource, but maybe it would be cleaner to make sure that only one pod applies the migration, potentially by switching from deployment to statefulset where pods have stable identifiers...? 🤔

olevski:
@ableuler you raise a good point about having more than one replica of the notebook service trying to do the migrations at the same time. Initially I thought this is not a problem. But, as you point out, it may be a problem. If we have to do more than one migration then it could be that one pod launches a migration from v1 -> v2 where as another has already done v2 -> v3. Then it may be possible to get in a situation where because of a race condition one user session gets migrated from v2 -> v3 very briefly only to be migrated back to v2 by some other pod. I think that using a statefulset for the notebook service is the cleanest solution. This way I can pass the pod name as an environment variable in the init container that does the migration. Then the first step of the migration will be to check if it is happening in the `"*-0"` pod. All other init containers will not do anything.

resource_migrations/migration_1.py
olevski: I removed the sleep.

resource_migrations/run_all.py
olevski: I added a migration version. This allows for a few things:
- migrations can run exactly on the right resources
- the migration version is part of the notebook helm chart - so now when new notebooks are created the migration version is already part of resource - so that the oldest migrations do not always run

resource_migrations/migration_1.py
olevski: I am using pagination now.

olevski:
So now using the migration labels the idea is that:
- the first migration selects all resources that are fully missing the migration label (had no clue that k8s allows you to select resources with the labels like this but it does)
- the second and subsequent migrations will select based on the labels of the `n-1` migration so that they truly get only the resources that need the migrations

helm-chart/renku-notebooks/values.yaml
```diff
@@ -26,6 +26,9 @@ userSessionPersistentVolumes:
   ## above is set to false.
   useEmptyDirSizeLimit: false
 
+migrations:
```
ableuler, line 29: Can one actually set this here or is the migration version strictly determined by the image of the notebooks service application? Also, I know I suggested "migration version", but maybe "schema version" is actually better. But that might trigger the false expectation of a relational DB... not sure 🤔

olevski: I will rename this to "schema version".

olevski: And actually the current migration version can simply be hardcoded in the config for the notebook service - and taken out of the values file. This is not needed in the migrations initContainer. The migrations initContainer will run all the migrations that are available/added. I dont see any point in trying to determine how many/which migrations to run in the initContainer.

I just have to remember to bump the current migration version in the config file when we add migrations.

ableuler (dismissed):

olevski:
@ableuler I am sorry about dismissing the approval. There was just a small issue with the black formatting checks after I did the renaming from migrations to schema. Can you re-approve please?

ableuler (approved):

PR #887, opened 2022-01-25
feat(app): return logs from all containers

### LINKED ISSUE #869: return logs from all containers in the session

olevski:
/deploy #persist

closes #869


renku_notebooks/api/classes/server.py
```diff
@@ -975,24 +975,38 @@ def stop(self, forced=False):
...
+                    self._k8s_namespace,
+                    container=container_name,
+                    tail_lines=max_log_lines if max_log_lines > 0 else None,
+                    timestamps=True,
+                )
+            except ApiException as err:
+                if err.status in [400, 404]:
+                    continue  # container does not exist or is not ready yet
+                else:
+                    raise err
```
Panaetius, line 1002:
```suggestion
                    raise
```

That way the original stack trace is preserved.

Even better is `raise DedicatedExceptionThatLogCantBeReadForSomeReasonException() from err`, then you get two exceptions/stack traces, the original one followed by this follow up error (naming left up to the reader) that could add additional detail like `Couldn't read log from container XYZ`.

renku_notebooks/api/schemas.py
```diff
@@ -533,18 +533,29 @@ class ServerOptionsUI(Schema):
...
+class ServerLogs(_ServerLogs):
+    class Meta:
+        unknown = INCLUDE  # only affects loading, not dumping
+
+    @pre_dump
+    def split_by_newline(self, data, *args, **kwargs):
+        # the ui expects the logs to be a list of strings
+        # with one log line being a single element in the list
+        for key in data:
+            data[key] = str.splitlines(data[key])
```
Panaetius, line 550: Looping over and modifying structures at the same time isn't good practice, it's quite prone to bugs (even though it might work in this case).
```suggestion
        data = {k: v.splitlines() for k,v in data.items()}
```

renku_notebooks/api/schemas.py
```diff
@@ -533,18 +533,29 @@ class ServerOptionsUI(Schema):
...
+        return data
+
+    @post_dump(pass_original=True)
+    def keep_unknowns(self, output, orig, **kwargs):
+        for key in orig:
+            if key not in output:
+                output[key] = orig[key]
```
Panaetius, line 557:
```suggestion
        output = {**orig, **output}
```
is probably the simplest way to merge dicts.

Panaetius (changes requested):

renku_notebooks/api/classes/server.py
olevski: @Panaetius I will do this `raise DedicatedExceptionThatLogCantBeReadForSomeReasonException() from err` or something very similar in [another PR](https://github.com/SwissDataScienceCenter/renku-notebooks/pull/849) that is already open that is supposed to cleanup and standardize all the errors in the notebook service.

Panaetius (approved):
:ship:

olevski:
NOTE: This will break the UI. It should be merged together with the corresponding change on the UI side. This is the related issue on the UI side: https://github.com/SwissDataScienceCenter/renku-ui/issues/1647

Panaetius (approved):

Panaetius (approved):

olevski:
I am sorry @Panaetius the approval got dismissed because I had to resolve some merge conflicts. Can you re-approve please?

PR #893, opened 2022-01-28
chore(chart): remove hub configmap that is unused

olevski:
Found a leftover unused configmap for jupyterhub in the helm chart templates that is not needed or used anymore.


ableuler (approved):

olevski:
My bad we already took care of this in another PR.

PR #894, opened 2022-01-28
chore: fix swagger page

olevski:
/deploy #persist


olevski:
The only test that is failing in the acceptance tests is:
```
Scenario: User can launch Jupyter notebook when the project is private - done *** FAILED ***
[info]   The code passed to eventually never returned normally. Attempted 406 times over 2.0053905388 minutes. Last failure message: Connect to session button not found. (ProjectPage.scala:300)
```

I tested this on the CI deployment for this PR manually and it works. So this is most probably a random fluke.

rokroskar:
on the screenshot it looks like it's waiting for the server listing. But I don't see how this PR would have caused it

rokroskar (approved):

PR #908, opened 2022-02-03
chore: track release version in sentry

olevski:
According to the docs here: https://docs.sentry.io/platforms/python/configuration/releases/

Setting an environment variable like `SENTRY_RELEASE` will be picked up automatically by sentry and used to distinguish and track different release versions.

### DIFF
--- a/helm-chart/renku-notebooks/templates/statefulset.yaml
+++ b/helm-chart/renku-notebooks/templates/statefulset.yaml
@@ -108,6 +108,8 @@ spec:
             - name: SENTRY_SAMPLE_RATE
               value: {{ .Values.sentry.sampleRate | quote }}
             {{ end }}
+            - name: SENTRY_RELEASE
+              value: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
             - name: CERTIFICATES_IMAGE
               value: "{{ .Values.global.certificates.image.repository }}:{{ .Values.global.certificates.image.tag }}"
             - name: CUSTOM_CA_CERTS_SECRETS
--- a/helm-chart/renku-notebooks/templates/statefulset.yaml
+++ b/helm-chart/renku-notebooks/templates/statefulset.yaml
@@ -109,7 +109,7 @@ spec:
               value: {{ .Values.sentry.sampleRate | quote }}
             {{ end }}
             - name: SENTRY_RELEASE
-              value: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (commented):
Do you mind triggering a CI deployment to test it on the fly?

Just a comment here: in the UI, we use the `.Chart.Version` value (I guess `image.tag` is the same) for the sentry logs _without_ the repo. That keeps the version more compact since the reference repo is already defined as a unique project in Sentry.

P.S: this probably isn't useful, but we also ended up with a [function](https://github.com/SwissDataScienceCenter/renku-ui/blob/master/client/src/utils/helpers/sentry/Sentry.js#L182-L205) to unify the non-release versions with a `-dev` suffix. It makes it a bit easier to filter errors on the fly, but ofc it's also a loss of information.

Screenshot_20220204_092958

olevski:
@lorenzo-cavazzi I added the whole image name just in case someone builds a custom image - not one from the renku repo. This way I will know it. For example I push all my images to gitlab. And I know there are some users who have weird infra that requires them to rebuild the images and load them from a specific repo. I think it is useful to have all this info and I would prefer to keep this info rather than just a semver.

Also on the other hand when it comes to distinguishing between dev deployments I will just add a separate namespace for that. So all the CI deployments will go into a specific namespace - and I can filter them out like that. I honestly wasn't even planning on using sentry in the CI deployments.

olevski:
@lorenzo-cavazzi I will just go with the tag for the chart. It seems that the full image name (especially when it is on the gitlab registry) is really long and then it breaks sentry. I.e. it shows up in sentry but I cannot get any more release-specific info and clicking on the release in sentry returns a 404.

olevski:
Ok the releases (and errors caused on purpose) show up now.

This is deployed at https://tasko.dev.renku.ch. It still has the error in there on getting the logs.

You can see at: https://sentry.dev.renku.ch/organizations/sentry/projects/renku-notebooks/?project=5

lorenzo-cavazzi:
I just tried it from https://tasko.dev.renku.ch , but I get a `400` error on sentry APIs 
Screenshot_20220204_131437

olevski:
@lorenzo-cavazzi just use the swagger page. That is because I have the sentry keys for the ui enabled but dont have the right key. You can use an anonymous session. That is the easiest way so you dont have to log in at all.

Also that request for sentry must be coming from the ui and it fails because the key in my values file is out of date.

The requests to sentry from the notebook service will now show up in the browser at all.

PR #915, opened 2022-02-08
feat: add anti-affinity to stateful set

### LINKED ISSUE #820: Add anti-affinity to renku-notebooks pods
Autoscaling on the renku-notebooks pods works nicely to cover load spikes, but the setup isn't optimized yet for resilience. Specifically, we should set the minimum number of pods to 2 rather than 1 and also make sure that the pods belonging to this deployment get spread out. At the moment k8s will happily schedule them all on the same node if that makes sense from a resource perspective.

Panaetius:
closes #820 

/deploy #persist

### DIFF
--- a/helm-chart/renku-notebooks/templates/statefulset.yaml
+++ b/helm-chart/renku-notebooks/templates/statefulset.yaml
@@ -194,8 +194,17 @@ spec:
       nodeSelector:
 {{ toYaml . | indent 8 }}
     {{- end }}
-    {{- with .Values.affinity }}
       affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                      - {{ template "notebooks.name" . }}
+              topologyKey: "kubernetes.io/hostname"
+    {{- with .Values.affinity }}
 {{ toYaml . | indent 8 }}
     {{- end }}
     {{- with .Values.tolerations }}
--- a/helm-chart/renku-notebooks/templates/statefulset.yaml
+++ b/helm-chart/renku-notebooks/templates/statefulset.yaml
@@ -197,13 +197,15 @@ spec:
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
-            - labelSelector:
-                matchExpressions:
-                  - key: app
-                    operator: In
-                    values:
-                      - {{ template "notebooks.name" . }}
-              topologyKey: "kubernetes.io/hostname"
+            - weight: 100
+              podAffinityTerm:
+              - labelSelector:
+                  matchExpressions:
+                    - key: app
+                      operator: In
+                      values:
+                        - {{ template "notebooks.name" . }}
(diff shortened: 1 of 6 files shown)


olevski:
@Panaetius I checked and when the keys are duplicated then the one key later in the file is the one used. So the way that you have set things up things should work.

helm-chart/renku-notebooks/templates/statefulset.yaml
```diff
@@ -194,8 +194,17 @@ spec:
       nodeSelector:
 {{ toYaml . | indent 8 }}
     {{- end }}
-    {{- with .Values.affinity }}
       affinity:
+        podAntiAffinity:
```
olevski, line 198: This section should look like this:

```yaml
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 100
              podAffinityTerm:
                labelSelector:
                  matchExpressions:
                    - key: app
                      operator: In
                      values:
                        - {{ template "notebooks.name" . }}
                topologyKey: "kubernetes.io/hostname"
```

The `weight` field is mandatory. And you are missing the `podAffinityTerm` level.

olevski (changes requested):

helm-chart/renku-notebooks/templates/statefulset.yaml
olevski: p.s. can you also please tag with `/deploy #persist` after you make the changes?

Panaetius: Right I mixed up `preferredDuringSchedulingIgnoredDuringExecution` where it's required and `requiredDuringSchedulingIgnoredDuringExecution` where it's not.

Though if we should have a strong preference for it or require it is an open question

helm-chart/renku-notebooks/templates/statefulset.yaml
```diff
@@ -194,8 +194,19 @@ spec:
...
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 100
+              podAffinityTerm:
+              - labelSelector:
```
olevski, line 202:
```suggestion
                labelSelector:
```

olevski: @Panaetius pretty sure this is breaking the tests.

olevski (approved):
I can confirm in the CI deployment that this is working as expected:
```
renku-ci-nb-915-notebooks-0                            1/1     Running     0          2m35s   10.42.13.25    worker-dev.renku-fat-nodes-2   <none>           <none>
renku-ci-nb-915-notebooks-1                            1/1     Running     0          98s     10.42.3.27     worker-dev.renku-default-6     <none>           <none>
```

olevski:
I cannot replicate the previous failure of the integration tests. I think it was caused by some changes happening in the infra at the time where most of the nodes were scaled down a lot and it was hard to schedule things. Rerunning the tests cleared this.

In addition the acceptance tests fail on starting an anonymous session. I tested this in the CI deployment and things work as expected - so it is just a fluke. I can launch an anonymous sessions without a problem.

PR #898, opened 2022-02-01
feat(app): simplify server status

olevski:
/deploy renku-ui=1646-use-simplified-sessions-state renku=000-acceptance-tests #persist

closes #809 

!BREAKING CHANGE for API / UI!

PR #923, opened 2022-02-11
feat(app): better messages on session launch fail

### LINKED ISSUE #868: return meaningful messages when sessions launch fails
related to #809 

This should decipher the status codes and/or messages provided by the main pod and any containers and return reasonable summary of the problem.

It seems that the status codes can indeed be translated/used.

See:
- https://docs.docker.com/engine/reference/run/#exit-status
- https://tldp.org/LDP/abs/html/exitcodes.html
- https://medium.com/tailwinds-navigator/kubernetes-tip-how-to-disambiguate-a-pod-crash-to-application-or-to-kubernetes-platform-f6c1395a8d09

olevski:
This gives more useful messages when a session launch fails.

closes #868 

Includes changes in #898 - so it is better to review after #898 is reviewed and merged.

/deploy renku-ui=1646-use-simplified-sessions-state renku=000-acceptance-tests #persist


renku_notebooks/api/schemas.py
```diff
@@ -224,6 +224,14 @@ def resolve_gpu_fieldname(self, in_data, **kwargs):
         return in_data
 
 
+class ServerStatus(Schema):
+    state = fields.String(
+        required=True,
+        validate=lambda x: x in ["running", "starting", "stopping", "failed"],
```
Panaetius, line 230:
```suggestion
        validate=validate.OneOf(["running", "starting", "stopping", "failed"]),
```
I'm a big fan of using Enum for these kinds of things. Right now they're just disconnected strings in different parts of the file, but an Enum could make that safer and unify it. 
E.g.
```
class ServerStatusEnum(Enum):
    Running = "running"
    Starting = "starting"

    @classmethod
    def list(cls):
        return list(map(lambda c: c.value, cls))

[...]

validate=validate.OneOf(ServerStatusEnum.list())

[...]

return {
    "state": ServerStatusEnum.Running.value,
}
```

renku_notebooks/api/schemas.py
```diff
@@ -248,66 +256,133 @@ class Meta:
...
+                    return user_correctable_message
+            return (
+                f"There are failures in {num_failed_containers} auxiliary "
+                "server containers. Please restart your session as this may be "
+                "an intermittent problem. If issues persist contact your "
+                "administrator or the Renku team."
+            )
+
+        def get_all_container_statuses(js):
+            return js["status"].get("mainPod", {}).get("status", {}).get(
```
Panaetius, line 325: Not really something for this PR, but did we ever consider using something like JSONPath/XPath for getting nested dict values? I see a lot of .get(..., {}).get(..., {}) style calls in renku_notebooks and I think that could make a lot of things cleaner.

Like https://pypi.org/project/jmespath/ , https://pypi.org/project/dpath/ or https://github.com/kennknowles/python-jsonpath-rw (looks quite nice and is semi-active but hasn't [had an official release for a while](https://github.com/kennknowles/python-jsonpath-rw/issues/98) :disappointed:  )

Panaetius (commented):
Looks nice. Just two minor comments.

renku_notebooks/api/schemas.py
olevski: That is a great idea and I think we should use something like that. Because whenever I did not do `.get(..., {}).get(..., {})` I regretted it afterwards and had to fix bugs. I will open an issue.

olevski: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/945

renku_notebooks/api/schemas.py
olevski: This is really neat. Thanks I added it.

Panaetius (dismissed):

Panaetius (approved):

PR #938, opened 2022-02-17
feat(app): version endpoint

### LINKED ISSUE #879: Add /version endpoint
Renku notebooks should expose a public `GET /version` endpoint.

The response format should be the following one:
```
{
    "name": string,
    "versions": [
        {
            "version": string,
            "data": object
        }
    ]
}
```

* "name" will be `renku-notebooks`
* "version" should be a SemVer valid value, possibly the tag or some combination of last tag + commit short-sha if it's a dev version
* "data" is optional, and it can be any arbitrary valid JSON object.

mohammad-sdsc:
Fixes #879 

/deploy


renku_notebooks/api/__init__.py
```diff
@@ -30,4 +31,36 @@ def health():
...
+    """
+    from ..version import __version__
+
+    info = {
+        "name": "renku-notebooks",
+        "versions": [
+            {
+                "version": __version__,
+                "data": {
+                    "anonymous_sessions_enabled": config.ANONYMOUS_SESSIONS_ENABLED,
```
olevski, line 57:
```suggestion
                    "anonymousSessionsEnabled": config.ANONYMOUS_SESSIONS_ENABLED,
```

olevski: This is more in line with naming in the values file. And I would like to transition to returning camelCase in endpoints as much as possible. Whereas internal python code would be snake_case.

renku_notebooks/api/__init__.py
```diff
@@ -30,4 +31,36 @@ def health():
...
+                    "s3_mount_enabled": config.S3_MOUNTS_ENABLED,
```
olevski, line 58:
```suggestion
                    "cloudstorageS3Enabled": config.S3_MOUNTS_ENABLED,
```

olevski: ...same as above - this is a bit more consistent with existing terminology.

tests/integration/test_session_creation.py
```diff
@@ -19,13 +19,26 @@
...
 import os
+import semver
 
 
 def test_can_check_health():
     response = requests.get(os.environ["NOTEBOOKS_BASE_URL"] + "/health")
     assert response.status_code == 200
 
 
+def test_version_endpoint():
```
olevski, line 30: Thanks for adding this 🥇

olevski (changes requested):
Thanks Mohammad just a minor change in how we pull in the version. What you suggested is fine, but it is a bit less convoluted to pull this through Helm rather than parse the template when the image is built. Also we already pull in a lot of configuration settings from the values file like this.

olevski:
One more thing I just thought of Mohammad. Please remember the register the schema so that it properly works with swagger here: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/__init__.py#L142.

Dockerfile
```diff
@@ -17,6 +18,10 @@ COPY resource_schema_migrations /resource_schema_migrations
 # Set up the flask app
 ENV FLASK_APP=/renku-notebooks/renku-notebooks:create_app
 
+# Update version
+COPY helm-chart/renku-notebooks/Chart.yaml /renku-notebooks/
+RUN echo "__version__ = \"$(yaml-get --query .version Chart.yaml)\"" > renku_notebooks/version.py
```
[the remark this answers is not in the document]
mohammad-sdsc: Thanks! This is much cleaner.

renku_notebooks/api/notebooks.py
```diff
@@ -41,6 +42,36 @@
...
+              schema: VersionResponse
+    """
+    info = {
+        "name": "renku-notebooks",
+        "versions": [
+            {
+                "version": config.NOTEBOOKS_SERVICE_VERSION,
+                "data": {
+                    "anonymousSessionsEnabled": config.ANONYMOUS_SESSIONS_ENABLED,
+                    "cloudstorageEnabled": config.S3_MOUNTS_ENABLED,
```
olevski, line 67: @mohammad-sdsc this is supposed to have one more level of keys that indicate the type of storage.

```suggestion
                    "cloudstorageEnabled": {
                        "s3": config.S3_MOUNTS_ENABLED,
                    },
```

olevski (changes requested):

olevski (approved):

PR #951, opened 2022-02-24
fix(app): keep git-proxy alive on session shutdown

### LINKED ISSUE #920: Properly handle LFS files and long-stopping sessions when creating autosaves
Currently all (or at least some) of our containers are ignoring the `SIGTERM` signal from k8s. When this is the case k8s waits for a default duration of 30 secs and then sends the forceful termination `SIGKILL` signal. 

This is essentially preventing the creation of autosave branches in the case when creating such a branch takes more than 30 seconds. This leads to lost work - especially in cases where the repo has a lot of LFS files and it takes a bit of time for the autosave branch to be crated.

In addition to this we should also properly handle any large uncommitted files that should go to LFS and put them in LFS in the autosave branch. We can potentially use `renku save` for this purpose.

olevski:
We noticed that in many cases the autosave branch does not get created. This was especially true if the unsaved commits in the session has large data files.

The main reason for this is that k8s waits only 30 seconds after it has sent the termination signal (SIGTERM) to a pod before it forcefully shuts it down. So if the autosave takes more than 30 seconds to be created then it will be interrupted before it is fully pushed/created.

K8s has the `terminationGracePeriodSeconds` for this purpose. But for this to work properly all containers have to properly handle the initial SIGTERM signal that k8s sent. If not then the pod will always take `terminationGracePeriodSeconds` to shutdown.

By adding `tini` to containers now the SIGTERM signal is properly handled. But now the containers shut down so quickly that the autosave branch still cannot be created. 

This is solved by the following:
- the git proxy catches the SIGTERM event and waits until it receives a GET request on the health port at /shutdown
- the git proxy therefore shuts down only if it gets such a /shutdown request or the terminationGracePeriod expires
- the preStop step that creates the autosave branch now sends a GET request to /shutdown when it is done to indicate to the git proxy it can shutdown
- to avoid confusion and edge cases the git proxy does not always shut down when it receives a request at /shutdown. It will only properly act on the /shutdown url once it receives a SIGTERM, if this hasn't happened then calling /shutdown has no effect
- in addition to the above the code that creates the autosave finds any large files and checks them in LFS - prior to this change all files were simply checked in git
- in the case of anonymous sessions the git proxy will shutdown as soon as it receives the SIGTERM and will not wait for /shutdown request, this is because the anonymous sessions cannot create autosaves

closes #920


git_services/git_services/sidecar/rpc_server.py
```diff
@@ -53,40 +57,60 @@ def status(path: str = ".", **kwargs):
...
+        git_proxy_health_port = os.getenv("GIT_PROXY_HEALTH_PORT", "8081")
+        repo_path = os.environ.get("MOUNT_PATH")
+        status_result = status(path=repo_path)
+        should_commit = not status_result["clean"]
+        should_push = status_result["ahead"] > 0
 
-    initial_commit = os.environ["CI_COMMIT_SHA"][0:7]
-    current_commit = status_result["commit"][0:7]
-    current_branch = status_result["branch"]
+        if not should_commit and not should_push:
```
vigsterkr, line 67:
```suggestion
        if not (should_commit or should_push):
```

vigsterkr (approved):
🚀

vigsterkr (dismissed):
:shipit:

PR #955, opened 2022-03-03
feat: move autosave to python

### LINKED ISSUE #883: Convert pre-stop autosave generating script form bash to python
The pre-stop hook that creates the autosave branches when a session is closed accidentally or a user forgets to commit/push their changes is a bash script defined in a configmap in the notebooks helm chart.

It would be beneficial to:
- covert this to a python script
- move it into its own folder with a dedicated Dockerfile similar to the git-clone script (they could even share the same docker file because both will most likely only need the gitlab python library)
- write some tests for this code

In addition to the above, the script can also figure out and assign a proper initial commit SHA. 

Currently the autosave branches are named like this: `renku/autosave/$RENKU_USERNAME/${CURRENT_BRANCH}/${INITIAL_SHA}/${LOCAL_SHA}` where `INITIAL_SHA` is the commit SHA from which the session was launched. However if the user commits and pushes something and then has other unsaved work the `INITIAL_SHA` should be the last pushed commit not the commit from which the session is launched. The python code in the pre-stop hook can figure this out and appropriately name the autosave branch.

### LINKED ISSUE #946: Add sentry to preStop hook that creates autosaves

Panaetius:
closes #883 
closes #946 

/deploy #persist


git_services/git_services/sidecar/run_command.py
```diff
@@ -0,0 +1,21 @@
+import argparse
+from git_services.git_services.sidecar.config import config_from_env
```
olevski, line 2:
```suggestion
from git_services.sidecar.config import config_from_env
```

This is def a problem. But I saw this only after changing the `terminationGracePeriodSeconds` on the session pod. This causes everything to "hang" and I got a shell into the sidecar and tried to run the autosave command manually. This is actually how you can test things out without shutting down the session. I will see if this is the only problem.

Getting caught up on small things like this which are obvious only when you deploy something is unfortunately common in the notebook service IMO.

Panaetius: ah I already fixed that in the rpcserver.py but must have missed it here

git_services/git_services/sidecar/rpc_server.py
```diff
@@ -1,26 +1,90 @@
...
+    """Create an autosave branch with uncommitted work."""
+    repo_path = os.environ.get("MOUNT_PATH")
+    status_result = status(path=repo_path)
+    should_commit = not status_result["clean"]
+    should_push = status_result["ahead"] > 0
+
+    if not should_commit and not should_push:
+        return
+
+    initial_commit = os.environ["CI_COMMIT_SHA"][0:7]
```
olevski, line 64: This and the `RENKU_USERNAME` environment variable are missing from the rpc server. You should add these in the patch.

You can do so like this, in the patch that adds the git sidecar:
```
                           {
                                "name": "CI_COMMIT_SHA",
                                "value": f"/work/{server.commit_sha}",
                            },
                            {
                                "name": "RENKU_USERNAME",
                                "value": f"/work/{server._user.username}",
                            },
```

git_services/git_services/sidecar/rpc_server.py
```diff
@@ -1,26 +1,90 @@
...
+
+    if should_commit:
+        cli.git_add("-A")
+        cli.git_commit(
+            f"-m 'Auto-saving for {user} on branch {current_branch} from commit {initial_commit}'"
+        )
+
+    cli.git_push(f"origin {autosave_branch_name}")
+
+    cli.git_reset("--soft current_branch")
```
olevski, line 86:
```suggestion
    cli.git_reset(f"--soft {current_branch}")
```

git_services/git_services/sidecar/rpc_server.py
```diff
@@ -1,26 +1,90 @@
...
+    cli.git_branch(f"-D {autosave_branch_name}")
```
olevski, line 87:
```suggestion
    cli.git_checkout(current_branch)
    cli.git_branch(f"-D {autosave_branch_name}")
```

We need to switch to the original branch before we delete the autosave. Otherwise you cannot delete the branch you are currently on.

olevski (changes requested):
So I found three issues. And another issue in the `python-git-clone` branch. I fixed the issue in the `python-git-clone` branch. So if you make the requested changes and pull in the latest from `python-git-clone` then everything should be fine.

Issue #882, opened 2022-01-24
Convert git-clone script to python


olevski:
At almost 200 lines and some not so simple logic all implemented in bash I think it may be a good idea to use python here.

This way we can write some unit tests for this. And it should make things a bit more easier to understand.

Panaetius:
https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/git-clone/clone.sh

rokroskar:
done in https://github.com/SwissDataScienceCenter/renku-notebooks/pull/956

PR #956, opened 2022-03-03
feat(app): use python for git clone

### LINKED ISSUE #928: Git clone init container does not fail when it runs out of disk space
To reproduce:
1. Create a project and add some LFS files
2. Launch a session enabling automatic download of LFS files and selecting a storage option that is too low
3. Look at k8s logs - the git clone init container should have failed but it did not - instead the `jupyter-server` container failed when it tried to launch with a full hard drive

I think that the best way to address this is to convert the git-clone script form bash to python. That way one can capture and handle the exception that git throws when it runs out of space. Then the python script can exit with a specific exit code that will signal to the notebook service that the session is running out of disk space.

olevski:
@Panaetius here I also re-organized the code so that stuff can be shared between the git-clone and rpc server. Actaully I use the same python environment in both.

Let me know what you think. You can base your PR off of this.

closes #928 #882


git_services/git_services/init/cloner.py
```diff
@@ -0,0 +1,150 @@
...
+        self.cli.git_fetch(self.remote_name)
+        res = self.cli.git_checkout(branch)
+        _, err = res.communicate()
+        if res.returncode != 0 or len(err) != 0:
+            if b"no space left on device" in err:
+                # INFO: not enough disk space
+                raise errors.NoDiskSpaceError
+            else:
+                # INFO: the branch simply does not exist, create it and continue
+                self.cli.git_checkout(f"-b {branch}")
```
Panaetius, line 94: Shouldn't we raise an error if the branch doesn't exist and let the user know?

Or at least not create the branch and just continue on the default branch.

git_services/git_services/init/cloner.py
```diff
@@ -0,0 +1,150 @@
...
+        # INFO: Reset the file tree to the auto-saved state.
+        self.cli.git_reset(f"--hard {autosave_branch}")
+        # INFO: Reset HEAD to the last committed change prior to the autosave commit.
+        pre_save_local_commit_sha = autosave_items[7]
+        self.cli.git_reset(f"--soft {pre_save_local_commit_sha}")
+        # INFO: Unstage all modified files.
+        self.cli.git_reset("HEAD .")
+        # INFO: Delete the autosave branch both remotely and locally.
+        autosave_local_branch = autosave_items[2:]
+        self.cli.git_push(f'{self.remote_name} :"{autosave_local_branch}"')
```
Panaetius, line 127:
```suggestion
        self.cli.git_push(f'{self.remote_name} --delete "{autosave_local_branch}"')
```
https://git-scm.com/docs/git-push#Documentation/git-push.txt---delete Does the same but looks leaner/more readable

Panaetius (changes requested):

git_services/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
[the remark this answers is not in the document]
olevski: Oh I did not know this. That is neat. I will move it.

git_services/git_services/init/cloner.py
olevski: @Panaetius this is a good question. I was just directly translating what we do in the shell script. And there is [this](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/git-clone/clone.sh#L64) line in the shell script which does the same thing. I will ping Andreas as to why this is the case.

olevski: I raise an error now - that makes sense. I checked this with Andreas too.

Panaetius (changes requested):

git_services/git_services/cli/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+
+
+class RepoDirectoryDoesNotExistError(Exception):
+    pass
+
+
+class GitCLI:
```
aledegano, line 17: I'm looking at this from a 10000ft view so disregard my question if it's dumb: have you considered GitPython to interface with Git instead of going through the CLI?
https://gitpython.readthedocs.io/en/stable/tutorial.html

I have used it in the past and while sometime is not the easiest library to use, it still works well and quite reliably.

Panaetius: GitPython is not actively maintained anymore (just major bugfixes), with the maintainer sometimes just going "Yeah, don't use that feature, it's broken and we won't improve on it" to issues. At least in renku-python we really hope there is a good replacement for it in the future (pygit isn't really an option and dulwich isn't there yet).

Panaetius (dismissed):

git_services/git_services/cli/__init__.py
aledegano: ah, thanks for the explanation, I was out of the loop.
Is really a big hole in the Python libraries then :/

olevski: @aledegano yes Ralf mentioned GitPython. The main concern I had with it is that the project is in maintenance mode. It seems the author is working on other things and relying on the community to maintain it. But if there is a bug or something is wrong I am not so sure that this will be addressed that quickly.

olevski: I am sorry for my late reply. Somehow I completely missed the already present and very detailed explanations from Ralf. I think in my defence they weren't loaded on my page. But it could be that I just missed them.

aledegano: no worries, thank you both for the explanation.

olevski:
@Panaetius I messed up and when I merged you suggestions I forgot that the `Optional` stuff wasn't imported from typing. So the tests failed because of this.

Panaetius (approved):

PR #963, opened 2022-03-14
feat(app): add sentry to git clone

### LINKED ISSUE #376: Add sentry to git-clone init container
Quite a few things can go wrong in the init-container (such as #375), it would be great to have some monitoring of that.

olevski:
This should be reviewed and merged after #956 is merged.

This adds the possibility to add separate sentry trackers for the init container in every sessions that clones and for the sidecar  rpc server container which (among other things) creates the autosave.

Because these are completely different environments than the renku-notebooks flask server I added separate projects for them. And the values file enables people to add separate sentry dsn for each container. I think this makes sense rather than seeing everything lumped in one. This way the errors from the flask server and any number of sessions always have to be combined in the same sentry project. This way we have flexibility to make them the same (by providing the same dsn in all places) or not.

In addition it is possible that some people may not want to add sentry in user sessions but may want to do so in the notebook server.

I caused the autosave creation and cloning to fail artificially and the errors show up:
https://sentry.dev.renku.ch/organizations/sentry/issues/415/?project=10
https://sentry.dev.renku.ch/organizations/sentry/issues/414/?project=11

closes #376


helm-chart/renku-notebooks/templates/statefulset.yaml
```diff
@@ -106,6 +106,22 @@ spec:
               value: {{ .Values.sentry.sampleRate | quote }}
             - name: SENTRY_RELEASE
               value: {{ .Chart.Version | quote }}
+            - name: GIT_CLONE_SENTRY_ENABLED
```
olevski, line 109: @Panaetius we do not need double underscores here. This is just injected into the notebook service container.

The notebook service then uses these environment variables and patches them into the session. Once things are patched into the sessions they need to have the double underscores so that they can be parsed by dataconf.

Here I did not add any double underscores because we do not use dataconf at this level yet.

olevski: The failing integration tests were a fluke. I reran them locally too without a problem.

Panaetius (approved):

PR #968, opened 2022-03-17
feat(app): show session resource usage

### LINKED ISSUE #727: Provide information about hard drive space and usage
This issue has remnants of a general discussion about how to communicate disk usage to users. Anything that is older than February 2022 can be ignored.

Some of the issues raised here are addressed by recent PRs such as: #923.

The resource usage as well as requests should be returned in the /servers/<name> and /servers endpoints.

The response should use the following schema:
```python
class ResourceUsage(Schema):
    cpu = fields.Str(required=False)
    memory = fields.Str(required=False)
    storage = fields.Str(required=False)

class ResourceRequests(ResourceUsage):
    gpu = fields.Str(required=False)

    @pre_load
    def resolve_gpu_fieldname(self, in_data, **kwargs):
        if "nvidia.com/gpu" in in_data.keys():
            in_data["gpu"] = in_data.pop("nvidia.com/gpu")
        return in_data

class Resources(Schema):  # Simply will replace UserPodResources
    requests = fields.Nested(ResourceRequests(), required=True)
    usage = fields.Nested(ResourceRequests(), required=False)
```

The most current data on resource usage from the session can be extracted from the JupyterServer resource manifest at `.status.mainPod.resourceUsage`. Here is an example of the format of the field:

```yaml
    resourceUsage:
      cpuMillicores: 5.633821999999999
      disk:
        availableBytes: 1002876928
        totalBytes: 1006526464
        usedBytes: 3649536
      memoryBytes: 80433152
```

What is currently being returned by the notebook service is what was requested when the session was launched. With this we will return what was requested AND what is being used.

Breaking API change requires a follow up from the UI.

**NOTE**: You cannot assume that `.status.mainPod` will be there all the time and same goes for any nested fields. The JupyterServer can exist in the cluster with just `status` being blank or definitely not containing even `mainPod` when sessions are starting`. Also if the metrics server is not enabled in the k8s cluster where amalthea is running then the cpu usage will be blank. It is also possible that due to problems with the image that is running the sessions also other parameters are missing.

olevski:
closes #727 

NOTE to reviewers: 90% of the changes here are the result of moving files around - i.e. instead of having one large file that contains all the schemas I broke all of these up. I will annotate the PR with comments to indicate what is exactly new code. Anything that is not annotated as new code in the diff is actually old code that just got moved.

~TODO: Fix integration tests when there is agreement on the units and final format of responses with the UI.~

/deploy renku-ui=1791-servers-resources #persist


renku_notebooks/api/__init__.py
```diff
@@ -1,33 +0,0 @@
-# -*- coding: utf-8 -*-
```
olevski, line 1: all of this got moved into renku_notebooks/api/health.py

renku_notebooks/api/schemas/autosave.py
```diff
@@ -0,0 +1,28 @@
+from marshmallow import fields, pre_dump, Schema
```
olevski, line 1: Not new code - just result from splitting the single large schemas.py

renku_notebooks/api/schemas/cloud_storage.py
```diff
@@ -0,0 +1,33 @@
+from marshmallow import fields, EXCLUDE, validate, ValidationError, Schema, post_load
```
olevski, line 1: Not new code - just result from splitting the single large schemas.py

renku_notebooks/api/schemas/logs.py
```diff
@@ -0,0 +1,14 @@
+from marshmallow import Schema, INCLUDE, post_dump, fields
```
olevski, line 1: Not new code - just result from splitting the single large schemas.py

renku_notebooks/api/schemas/servers_post.py
```diff
@@ -0,0 +1,56 @@
+from marshmallow import fields, Schema, ValidationError, validates_schema
```
olevski, line 1: Not new code - just result from splitting the single large schemas.py

renku_notebooks/api/schemas/version.py
```diff
@@ -0,0 +1,24 @@
+from marshmallow import fields, Schema
```
olevski, line 1: Not new code - just result from splitting the single large schemas.py

olevski:
@lorenzo-cavazzi this is ready now but needs a change on the ui. Because the resources response from the api doesn't just send the requests but also requests and usage.

As we discussed the units are the same as before.

olevski:
The changes in the response are as follows:

OLD:
```
    annotations = fields.Nested(UserPodAnnotations())
    name = fields.Str()
    state = fields.Dict()
    started = fields.DateTime(format="iso", allow_none=True)
    status = fields.Nested(ServerStatus())
    url = fields.Str()
    resources = {"cpu":"1", "memory":"1G", }  <<<< Changes are only here see below
    image = fields.Str()
```

NEW:
```
    annotations = fields.Nested(UserPodAnnotations())
    name = fields.Str()
    state = fields.Dict()
    started = fields.DateTime(format="iso", allow_none=True)
    status = fields.Nested(ServerStatus())
    url = fields.Str()
    resources = { requests: {cpu:"1", memory:"1G"}, usage: {cpu:"0.25", memory:"0.50G"} }  <<< 
    image = fields.Str()
```

olevski:
@rokroskar yes we can. I have to resolve conflicts and bring it up to date.

olevski:
TODO: Ensure the code in the `schemas` folder is up to date with the current state of the schemas.

lorenzo-cavazzi (approved):
The code looks good to me! And the changes work as expected with SwissDataScienceCenter/renku-ui#1825

I believe this requires a review from a renku-notebook maintainer (i.e. @Panaetius , the only other person in the group :grin:  )

lorenzo-cavazzi (approved):
:rocket:

PR #969, opened 2022-03-17
fix(app): enable s3 flag in server options

olevski:
/deploy #persist

So this was breaking the ui and it was commented out when merged.

Now that the ui fully supports s3 buckets it should be enabled fully. I forgot to do this earlier when I did the 1.5.0 release of notebooks. It would have been ideal to do it then.


mohammad-sdsc (approved):
🚀

lorenzo-cavazzi (approved):

PR #990, opened 2022-04-01
feat(app): add endpoint to verify docker image availability

### LINKED ISSUE #989: Add a `GET check_docker_image` endpoint
We should expose a new endpoint `check_docker_image` to allow clients to use [the logic implemented in the `image_exists` function](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/util/check_image.py#L57) to verify if a docker image is available or not.

This endpoint should take a simple parameter `image_url` and return a simple JSON response
```
{
  available: <bool>
}
```

lorenzo-cavazzi:
This PR adds a new endpoint to verify the availability of any remote Docker image in a v2 registry `GET /images?image_url=<registry-image_url>`.

I believe importing the `utils` function directly in the route is not the best strategy, but I wasn't sure if it's worth it to make an `Image` Class just for that.
Feel free to adapt the PR in any way that makes sense :grin: 

Screenshot_20220401_163956

fix #989


olevski:
@lorenzo-cavazzi this is great. Thanks for doing this. I will just add a marshmallow schema for the request too. We use a plugin that allows us to do this. And with this you can get a 422 response when the request is not valid "for free". After that will approve and merge.

olevski (approved):

olevski:
A few additional changes here:
- the response body is always blank because the status codes are enough to indicate if the image exists (200) or it does not (404)
- relying on the status codes allows us to remove the schema definition
- added a simple integration test for the endpoint

PR #993, opened 2022-04-04
feat(app): switch git proxy to golang

olevski:
Found a bit better maintained proxy library in Go. This will resolve all the critical security vulnerabilities we have with the current node js proxy.

I also added tests for the proxy that ensure that the Git credentials are added in the right circumstances and edited the CI pipeline so that these tests run on every push.


olevski:
The integration tests are failing because the git-proxy shuts down before the autosave can be created when a session is stopped. This is exactly the problem that #951 will address. Therefore to keep the diffs sane I think that we should merge this PR first and then rebase #951 to work with this version of the git proxy.

This occurs more rarely in practice (but it still does) because the old version of the git proxy does not handle the SIGTERM signal properly and waits until it receives SIGKILL from k8s 30 seconds later. This is enough in most cases to pass the integration tests.

olevski:
The integration tests fully pass in #951. So this can be merged and then #951 can be reviewed and merged. Or we can just close this PR and merge just #951.

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+	"strings"
+
+	"github.com/elazarl/goproxy"
+)
+
+func main() {
+	config := parseEnv()
+	proxyHandler := getProxyHandler(config)
+	proxyServer := http.Server{
+		Addr:    fmt.Sprintf(":%s", config.ProxyPort),
```
vigsterkr, line 21: i'm wondering why not "0.0.0.0:%s" ?

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+		Handler: proxyHandler,
+	}
+	healthHandler := getHealthHandler(config)
+	healthServer := http.Server{
+		Addr:    fmt.Sprintf(":%s", config.HealthPort),
+		Handler: healthHandler,
+	}
+	go func() {
+		// Run the health server in the "background"
+		log.Printf("Health server active on port %s\n", config.HealthPort)
```
vigsterkr, line 31:
```suggestion
		log.Println("Health server active on port, config.HealthPort)
```

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+		log.Fatalln(healthServer.ListenAndServe())
+	}()
+	log.Printf("Git proxy active on port %s\n", config.ProxyPort)
+	log.Printf("Repo Url: %v, anonymous session: %v\n", config.RepoUrl, config.AnonymousSession)
```
vigsterkr, line 35:
```suggestion
	log.Println("Git proxy active on port", config.ProxyPort)
	log.Println("Repo Url:" config.RepoUrl, "anonymous session:", config.AnonymousSession)
```
as `%v` is anyways default formatting, or?

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+		validGitRequest = r.URL.Scheme == config.RepoUrl.Scheme &&
+			hostsMatch(r.URL, config.RepoUrl) &&
+			getPort(r.URL) == getPort(config.RepoUrl) &&
+			strings.HasPrefix(strings.TrimLeft(r.URL.Path, "/"), strings.TrimLeft(config.RepoUrl.Path, "/"))
+		if config.AnonymousSession {
+			log.Print("Anonymous session, not adding auth headers, letting request through without adding auth headers.\n")
+			return r, nil
+		}
+		if !validGitRequest {
+			log.Printf("The request %v does not match the git repository %v, letting request through without adding auth headers\n", r.URL, config.RepoUrl)
```
vigsterkr, line 133:
```suggestion
			log.Println("The request", r.URL", "does not match the git repository", config.RepoUrl, ", letting request through without adding auth headers")
```

vigsterkr: same `%v` is default formatting... afaik println is using that.

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+			return r, nil
+		}
+		log.Printf("Adding auth header to request: %v\n", r.URL)
```
vigsterkr, line 136:
```suggestion
		log.Println("Adding auth header to request:", r.URL)
```

git-https-proxy/main.go
```diff
@@ -0,0 +1,190 @@
...
+	})
+	handler.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+		proxyUrl, err := url.Parse(fmt.Sprintf("http://localhost:%s", config.ProxyPort))
+		if err != nil {
+			log.Fatalln(err)
+		}
+		client := &http.Client{Transport: &http.Transport{Proxy: http.ProxyURL(proxyUrl)}}
+		resp, err := client.Get(fmt.Sprintf("http://localhost:%s/ping", config.HealthPort))
+		if err != nil {
+			log.Printf("The GET request to /ping from within /health failed with: %v", err)
```
vigsterkr, line 178:
```suggestion
			log.Println("The GET request to /ping from within /health failed with:", err)
```

vigsterkr (approved):
some total nitpicking, nothing major. feel free to ignore those

git-https-proxy/main.go
olevski: That is a good point to make this explicit. I think that should work.

olevski: Did it in both places.

If someone is here and wondering why not `127.0.0.1` it is because the k8s probes do not originate from localhost. But from another IP. So if I listen only on `127.0.0.1` then the probes fail.

And the probes are setup to go through the proxy to hit the health endpoint. So the proxy definitely cannot listen only on `127.0.0.1`

vigsterkr (approved):
:shipit:

PR #1006, opened 2022-04-06
fix(app): handle user names in git clone

olevski:
Fixes a problem with names with a middle name during the setup of the repo when launching a session.

This was introduced by the conversion of the "git clone" code from bash to python.

### DIFF
--- a/git_services/git_services/init/cloner.py
+++ b/git_services/git_services/init/cloner.py
@@ -63,9 +63,9 @@ class GitCloner:
         self.cli.git_init()
         # NOTE: For anonymous sessions email and name are not known for the user
         if self.user.email is not None:
-            self.cli.git_config(f"user.email {self.user.email}")
+            self.cli.git_config(f"user.email '{self.user.email}'")
         if self.user.full_name is not None:
-            self.cli.git_config(f"user.name {self.user.full_name}")
+            self.cli.git_config(f"user.name '{self.user.full_name}'")
         self.cli.git_config("push.default simple")
 
     def _setup_proxy(self):


olevski:
I was able to reproduce this by adding a middle name to my account. With the fix on my personal deployment it works. But on dev things fail.

git_services/git_services/init/cloner.py
```diff
@@ -63,9 +65,9 @@ def _initialize_repo(self):
         self.cli.git_init()
         # NOTE: For anonymous sessions email and name are not known for the user
         if self.user.email is not None:
-            self.cli.git_config(f"user.email {self.user.email}")
+            self.cli.git_config(f"user.email '{self.user.email}'")
         if self.user.full_name is not None:
-            self.cli.git_config(f"user.name {self.user.full_name}")
+            self.cli.git_config(f"user.name '{self.user.full_name}'")
```
rokroskar, line 70: Should these be escaped double ticks instead? Not sure

olevski: I dont think so. I already tested it and it works. And the ticks do not show up in the git config. I think double ticks vs single only matters if you want the shell to substitute a variable in there and then you would use double ticks. But that is not the case here.

I can switch to escaped double ticks though if you think that helps.

olevski: I will open another PR with escaped double ticks. And you can approve either/or.

olevski: @rokroskar here is the PR with escaped double quotes. That works too.

olevski:
The integration tests are failing because gitlab closes the connection. This happens often lately.

git_services/git_services/init/cloner.py
olevski: Ok so the single of double quotes do not matter because all of these strings passed to `git_cli` are passed to `shlex.split` (https://docs.python.org/3.8/library/shlex.html#shlex.split) before being run with `subprocess.Popen`.

And `shlex.split` does not care whether you use single or double quotes. In both cases it properly figures out what is an argument and what isn't.

```
>>> shlex.split("git config user.name \"Some complicated name\"")
['git', 'config', 'user.name', 'Some complicated name']
>>> shlex.split("git config user.name 'Some complicated name'")
['git', 'config', 'user.name', 'Some complicated name']
```

But not using any quotes like before is definitely wrong:
```
>>> shlex.split("git config user.name Some complicated name")
['git', 'config', 'user.name', 'Some', 'complicated', 'name']
```

rokroskar: Ok thanks for checking

git_services/git_services/init/cloner.py
```diff
@@ -153,8 +155,12 @@ def _repo_exists(self):
     def run(self, recover_autosave, session_branch, root_commit_sha):
         logging.info("Checking if the repo already exists.")
         if self._repo_exists():
-            logging.info("The repo already exists - exiting.")
-            return
+            # NOTE: We expect a clean state, if something is left it means the git-clone
```
rokroskar, line 158: Is that true? What about the case when the pv is left over because something crashed?

olevski: Oh that is a good point. Yeah I will remove this part.

rokroskar (approved):

PR #1007, opened 2022-04-06
fix(app): handle user names in git clone

olevski:
Same as #1006  but uses double quotes to handle a name instead of single.


olevski:
Tested on my personal deployment. It works.

PR #1020, opened 2022-04-13
chore: run as non-root

### LINKED ISSUE #998: Run notebook service as non-root

olevski:
Ensures in the session and notebooks all containers run as non root.

closes #998 #999

/deploy #persist extra-values=notebooks.securityContext.enabled=null


git-https-proxy/Dockerfile
```diff
@@ -5,4 +5,5 @@ RUN go build -o /git-http-proxy main.go
 
 FROM alpine:3.15
 COPY --from=builder /git-http-proxy /git-http-proxy
+USER 1000:100
```
Panaetius, line 8:
```suggestion
USER 1000:1000
```
?

renku_notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -20,6 +20,12 @@ def main(server):
                     "path": "/statefulset/spec/template/spec/containers/-",
                     "value": {
                         "image": current_app.config["GIT_HTTPS_PROXY_IMAGE"],
+                        "securityContext": {
+                            "fsGroup": 100,
+                            "runAsGroup": 100,
```
Panaetius, line 25: 1000?

Panaetius (commented):
I think this is on purpose and not a typo, but shouldn't we be consistent with 1000:1000 ?

renku_notebooks/api/amalthea_patches/git_proxy.py
olevski: The jupyterlab images run with fsgroup/group 100. Let me see if we can switch fully to 1000.

olevski: It works.

Panaetius (approved):

PR #1031, opened 2022-04-21
chore: set runAsNonRoot to true

olevski:
This further locks down the security context.

/deploy #persist


Panaetius (approved):
Thank you!

PR #1035, opened 2022-04-25
fix(app): eliminate unnecessary gitlab calls

olevski:
There are unnecessary calls to the gitlab api when servers are instantiated.

In addition some properties can be cached so that they are quicker when they are called more than once.

/deploy

### DIFF
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -1,4 +1,5 @@
 from flask import current_app
+from functools import lru_cache
 import gitlab
 from itertools import chain
 from kubernetes import client
@@ -68,13 +69,7 @@ class UserServer:
         self.is_image_private = None
         self.image_workdir = None
         self.cloudstorage = cloudstorage
-        try:
-            self.gl_project = self._user.get_renku_project(
-                f"{self.namespace}/{self.project}"
-            )
-        except Exception as err:
-            current_app.logger.warning("Cannot find project because:", err)
-            self.gl_project = None
+        self.gl_project_name = f"{self.namespace}/{self.project}"
         self.js = None
 
     def _check_flask_config(self):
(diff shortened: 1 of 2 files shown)


renku_notebooks/api/classes/server.py
```diff
@@ -68,13 +69,7 @@ def __init__(
...
         self.image_workdir = None
         self.cloudstorage = cloudstorage
-        try:
-            self.gl_project = self._user.get_renku_project(
-                f"{self.namespace}/{self.project}"
-            )
-        except Exception as err:
-            current_app.logger.warning("Cannot find project because:", err)
-            self.gl_project = None
+        self.gl_project_name = f"{self.namespace}/{self.project}"
```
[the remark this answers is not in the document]
olevski: Previously I fetched and defined the gitlab project here. And if the project did not exist (which is possible) this would fail. Now I check for the gitlab project only during creation when I call the property (which is cached).

renku_notebooks/api/classes/server.py
```diff
@@ -136,27 +137,25 @@ def _branch_exists(self):
...
-            ).commits.get(self.commit_sha)
-        except Exception:
-            return False
-        else:
-            return True
+        if self.commit_sha is not None and self.gl_project is not None:
+            try:
+                self.gl_project.commits.get(self.commit_sha)
+            except Exception:
+                pass
```
rokroskar, line 155: at least log a warning?

renku_notebooks/api/classes/server.py
```diff
@@ -136,27 +137,25 @@ def _branch_exists(self):
...
-        if self.branch is not None:
+        if self.branch is not None and self.gl_project is not None:
             try:
-                self._user.get_renku_project(
-                    f"{self.namespace}/{self.project}"
-                ).branches.get(self.branch)
+                self.gl_project.branches.get(self.branch)
             except Exception:
-                return False
+                pass
```
rokroskar, line 144: log a warning?

rokroskar (dismissed):
Just a small comment to maybe log a warning if a branch/commit is requested that doesn't exist. Feel free to ignore if there's a reason for not doing that.

olevski:
@rokroskar I added the warnings you requested. Can you re-approve please?

rokroskar (approved):

PR #1047, opened 2022-05-02
chore(docs): add sequence diagram to readme

olevski:
@rokroskar here is a very high level sequence diagram.

Let me know what you think. Also in reality the notebook service makes some additional calls to Gitlab that will need to be removed. I already removed some in a recent PR - but there are probably a few left from the autosave endpoints.


rokroskar:
Thanks @olevski this looks great! I'm just trying to figure out how we could include this in the main renku docs without duplication? Is it possible to put the diagram in a file and link to it from the markdown?

olevski:
@rokroskar I think in this case the easiest is to create a separate markdown file with the sequence diagram and have a link to it in the README.md. Then the renku docs can use this file somewhow?

The other option is (if possible) to render/convert the markdown README.md from the notebook service repo and include it the renku read the docs. Do you think such a thing is possible?

olevski:
I think that may be the nicest way to do this. So then the docs are alwayssynced with what is in the notebooks. I can submit a PR for this in the Renku repo.

rokroskar:
This is the doc for markdown: https://www.sphinx-doc.org/en/master/usage/markdown.html

I've never tried it... what we do with the rst files which I think is not possible with markdown is to include only sections of them in the docs. That way you can e.g. cut out a sub-section of a readme to include in the doc pages. It won't be possible here with markdown I don't think.

rokroskar (approved):
Thanks!!

PR #1055, opened 2022-05-10
feat: add git error codes to notebooks response

### LINKED ISSUE #1037: Sync exit codes from git-clone with notebooks api
The updated git-clone code uses specific exit codes depending on the type of error that occurs.

The notebook service can utilize this to relay more meaningful messages to the ui.

Here are the different exit codes that occur when a repo is cloned: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/git_services/git_services/init/errors.py

This should be used in the notebook service here: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/api/schemas.py#L286

Panaetius:
closes #1037 

/deploy #persist


olevski:
@Panaetius this looks good to me. I just want to see that the message will really show up when you try to clone but run out of space. If it does not show up this is not really the fault of this PR though. But still it would be nice to double-check.

olevski:
@Panaetius as I suspected there were some issues in the code that are not part of this PR and were there from before that were preventing the error messages from properly surfacing.

I corrected these and I will submit a 2nd PR that will make the necessary changes into this branch. Then we can merge my changes into your branch and merge this branch into master. 

Or I can directly commit these changes into your feature branch? If you prefer this let me know and I will close my PR and commit directly here.

olevski (approved):

PR #1072, opened 2022-05-23
chore: amend changelog

olevski:
Adds an explanation about potential problems that may occur with the helm chart if an old section of the values.yaml file is present.

I already updated this in the release notes on github.


rokroskar (approved):
We should amend this also in the main renku changelog for the admins that won't dig this deep...

PR #1083, opened 2022-05-25
fix(app): s3 bucket location

### LINKED ISSUE #1044: Missing location in s3 bucket causes trouble
When launching a session with s3 buckets the endpoint in some cases can include a zone (especially for aws) or not.

If the zone is excluded the library assumes the zone is us-east-1. However if the zone is not specified using the python boto3 library will still properly find the bucket (whatever zone it is in). But the datashim library that mounts the bucket in k8s has issues if the zone is not specified in the endpoint and the bucket is actually not located in the us-east-1 zone.

We can examine the output from `head_bucket` to determine what is the zone of the bucket. If the `x-amz-bucket-region` field is found in the output, and it is not us-east-1 then we can check and make sure that it is included in the endpoint. If the location is not included we can fail the session start.

There is a `get_location` method that can be used. But this is not usually allowed to be called on most public buckets.

```
>>> client.head_bucket(Bucket="ursa-labs-taxi-data")
{'ResponseMetadata': {'RequestId': '5YQTB5441ENS5H6R', 'HostId': 'shlAXICd7pBp7ZhOAT/fV1GCiJ8gRs6BmUlBWJDiOxiMw33JpsVEYwEAQnkOlw7gpAR1NqinwdQ=', 'HTTPStatusCode': 200, 'HTTPHeaders': {'x-amz-id-2': 'shlAXICd7pBp7ZhOAT/fV1GCiJ8gRs6BmUlBWJDiOxiMw33JpsVEYwEAQnkOlw7gpAR1NqinwdQ=', 'x-amz-request-id': '5YQTB5441ENS5H6R', 'date': 'Thu, 28 Apr 2022 09:28:39 GMT', 'x-amz-bucket-region': 'us-east-2', 'x-amz-access-point-alias': 'false', 'content-type': 'application/xml', 'server': 'AmazonS3'}, 'RetryAttempts': 0}}
```

olevski:
This fixes two problems we have recently seen with s3 buckets:

1. Boto3 can handle a missing location but datashim (that actually mounts the buckets) cannot. Boto3 is used to check if the bucket exists before things are handed over to datashim. That is why we did not initially catch this.
This is the issue reported in #1044. In there for example mounting the bucket `ursa-labs-taxi-data` located in `http://s3.us-east-2.amazonaws.com` would fail if the endpoint was specified as `http://s3.amazonaws.com`. But this failure would occur in datashim and the API would not know about it. With this fix the API can catch these cases and not launch the session. The reason for this is that if the location is not specified in the url then the default is `us-east-1`. But when the location is omitted from the url and the bucket location is actually not `us-east-1` then datashim could not mount the bucket.

2. The endpoint has to be a regular http:// or https:// url. Passing in the hostname only or something else that is malformed does not. Marshmallow (which we already use for validation) already provides a field like this so we get the checks for free. Before any string was allowed.

closes #1044

### DIFF
--- a/renku_notebooks/api/classes/s3mount.py
+++ b/renku_notebooks/api/classes/s3mount.py
@@ -126,14 +126,26 @@ class S3mount:
     @property
     def bucket_exists(self):
         try:
-            self.client.head_bucket(Bucket=self.bucket)
+            res = self.client.head_bucket(Bucket=self.bucket)
         except (ClientError, EndpointConnectionError, NoCredentialsError, ValueError):
             current_app.logger.warning(
                 f"Failed to confirm bucket {self.bucket} for endpoint {self.endpoint} exists"
             )
             return False
-        else:
-            return True
+        # INFO: If the region is not the default (us-east-1) and it is not speicifed explicitly in
+        # the endpoint then datashim has trouble mounting the bucket even though boto can find it.
+        amz_bucket_region = (
+            res.get("ResponseMetadata", {})
+            .get("HTTPHeaders", {})
+            .get("x-amz-bucket-region")
+        )
+        if (
+            amz_bucket_region
+            and amz_bucket_region != "us-east-1"
+            and amz_bucket_region not in self.endpoint
+        ):
+            return False
+        return True
 
     @classmethod
     def s3mounts_from_js(cls, js):
--- a/renku_notebooks/api/schemas.py
+++ b/renku_notebooks/api/schemas.py
@@ -78,7 +78,9 @@ class LaunchNotebookRequestS3mount(Schema):
 
     access_key = fields.Str(required=False, missing=None)
     secret_key = fields.Str(required=False, missing=None)
-    endpoint = fields.Str(required=True, validate=validate.Length(min=1))
+    endpoint = fields.Url(
+        required=True, schemes=["http", "https"], relative=False, require_tld=True
+    )
     bucket = fields.Str(required=True, validate=validate.Length(min=1))
 
     @post_load


renku_notebooks/api/classes/s3mount.py
```diff
@@ -126,14 +126,26 @@ def get_manifest_patches(
...
-            self.client.head_bucket(Bucket=self.bucket)
+            res = self.client.head_bucket(Bucket=self.bucket)
         except (ClientError, EndpointConnectionError, NoCredentialsError, ValueError):
             current_app.logger.warning(
                 f"Failed to confirm bucket {self.bucket} for endpoint {self.endpoint} exists"
             )
             return False
-        else:
-            return True
+        # INFO: If the region is not the default (us-east-1) and it is not speicifed explicitly in
```
rokroskar, line 135:
```suggestion
        # INFO: If the region is not the default (us-east-1) and it is not specified explicitly in
```

renku_notebooks/api/classes/s3mount.py
```diff
@@ -126,14 +126,26 @@ def get_manifest_patches(
...
+        # the endpoint then datashim has trouble mounting the bucket even though boto can find it.
+        amz_bucket_region = (
+            res.get("ResponseMetadata", {})
+            .get("HTTPHeaders", {})
+            .get("x-amz-bucket-region")
+        )
+        if (
+            amz_bucket_region
+            and amz_bucket_region != "us-east-1"
+            and amz_bucket_region not in self.endpoint
```
rokroskar, line 145: why do we fail here? If we can determine the true location of the bucket, shouldn't we just go ahead (like boto does)? I guess that would mean amending the endpoint with the actual zone.

rokroskar (changes requested):

renku_notebooks/api/classes/s3mount.py
olevski: To me it seemed a bit risky and unexpected (from the users perspective) to change a request parameter that came directly from the user. But yes we can do this if we hardcode a dictionary of regions mapped to the correct endpoints.

The problem is that if the urls for these endpoints change then things will break and we will have to update the code.

olevski: Ok I will change the endpoint in this case then

olevski:
@rokroskar I tested this in my personal deployment. Now it does not matter which region for aws S3 you specify - as long as you specify a valid region. The notebook service will then correct and set the s3 bucket region correctly. So now you can launch a session by even setting the endpoint to `http://s3.us-west-2.amazonaws.com` and requesting the `ursa-labs-taxi-data` bucket. Even though the bucket is actually located in `us-east-2`. If the bucket does not exist then the proper message is returned. Also I tested that things also work if the general url like `http://s3.amazonaws.com` is provided regardless of wether the bucket is in the `us-east-1` region or not.

rokroskar:
@olevski I'm not sure overwriting a region that has been explicitly specified is what we should do - shouldn't we only "fix" the URL if the region is generic, i.e. just s3.amazonaws.com.

olevski:
@rokroskar you are right. I modified the code so that the endpoint is only modified with the location only if the general endpoint without location is specified initially in the request.

In other cases the requests fail if the location is specified but the bucket resides in a different location.

rokroskar:
Great, I _think_ that's the correct behavior... :)

PR #1124, opened 2022-06-23
fix(app): handle endpoints for s3 not on AWS

olevski:
This logic was changed recently to recognize the amazon aws s3 region. It was done under the assumption that this `x-amz-bucket-region` would only show up for AWS s3 endpoints and not in other cases. But this is not true. This field can be present even if the bucket is not hosted on AWS. Because of this some buckets that truly exist and could be reached were deemed to not exist by the notebook service.

This corrects those problems.

/deploy #persist


Panaetius (approved):
🚀

PR #1131, opened 2022-06-27
feat: add support for passing env variables to notebooks

### LINKED ISSUE #1015: Inject environment variables in session when launching it
We could add an additional field to the POST /notebooks/servers to add environment variables. 

Something like

```
{
  "branch": "master",
  "env": [{"name": "ENV1", "value": "1234"}]
}
```

The smartsky project is launching renku sessions from their web portal. They need a user token for their portal inside the renku session - right now users have to copy/paste it into their notebooks but it could be passed in as an environment variable automatically on launch.

Panaetius:
closes #1015 

Most of the changes are adding typing in some places and sorting imports.

### DIFF
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -593,12 +593,11 @@ class UserServer:
             for patch in patches.get("patch", []):
                 if (
                     patch.get("path")
-                    == "/statefulset/spec/template/spec/containers/0/env/-"
-                    and patch.get("value", {}).get("name") == "GIT_AUTOSAVE"
+                    == "/statefulset/spec/template/spec/initContainers/-"
                 ):
-                    server_options["lfs_auto_fetch"] = (
-                        patch.get("value", {}).get("value") == "1"
-                    )
+                    for env in patch.get("value", {}).get("env", []):
+                        if env.get("name") == "GIT_CLONE_LFS_AUTO_FETCH":
+                            server_options["lfs_auto_fetch"] = env.get("value") == "1"
         return {
             **current_app.config["SERVER_OPTIONS_DEFAULTS"],
             **server_options,
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -1,37 +1,36 @@
-from flask import current_app
+import base64
+import json
 from functools import lru_cache
-import gitlab
 from itertools import chain
+from typing import Any, Dict, List, Optional
+from urllib.parse import urljoin, urlparse
+
+import gitlab
+from flask import current_app
 from kubernetes import client
 from kubernetes.client.exceptions import ApiException
 from kubernetes.client.models import V1DeleteOptions
-import base64
-import json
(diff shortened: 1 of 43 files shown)


renku_notebooks/api/classes/server.py
```diff
@@ -602,6 +604,33 @@ def _get_server_options_from_js(js):
...
+            "MOUNT_PATH",
+            "PROJECT_NAME",
+            "GIT_CLONE_REPO",
+        ]
+
+        for patches in js["spec"]["patches"]:
+            for patch in patches.get("patch", []):
+                if (
+                    patch.get("path")
+                    == "statefulset/spec/template/spec/containers/0/env/-"
```
olevski, line 625:
```suggestion
                    == "/statefulset/spec/template/spec/containers/0/env/-"
```

tests/unit/test_server_class/test_manifest.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.mark.parametrize(
+    "parameters,expected",
+    [
+        (
+            {"environment_variables": {"TEST": "testval"}},
+            "/containers/0/env/-', 'value': {'name': 'TEST', 'value': 'testval'}",
+        )
+    ],
+)
+def test_session_manifest(
```
olevski, line 15: This is a really nice way to test a decent portion of the server class functionality. 

It would be nice to add a few lines (or a separate test) to check the opposite works. I.e. that you can call `UserServer.from_js(user, manifest)` and get the environment variables in the `.environment_variables` on the server instance. 

This is how I quickly attempted to modify the test for this case and how I caught the missing `/` from the other comment. But I think you made this function to be more general based on the naming. So we can either rename the function to only test the patching and parsing env variables.


```python
@pytest.mark.parametrize(
    "parameters,expected",
    [
        (
            {"environment_variables": {"TEST": "testval"}},
            "/containers/0/env/-', 'value': {'name': 'TEST', 'value': 'testval'}",
        )
    ],
)
def test_session_manifest(
    parameters, expected, patch_user_server, user_with_project_path, app
):
    """Test that session manifest can be created correctly."""

    user = user_with_project_path("namespace/project")
    with app.app_context():
        base_parameters = {
            "user": user,
            "namespace": "test-namespace",
            "project": "test-project",
            "image": None,
            "server_options": {
                "lfs_auto_fetch": 0,
                "defaultUrl": "/lab",
                "cpu_request": "100",
                "mem_request": "100",
            },
            "branch": "master",
            "commit_sha": "abcdefg123456789",
            "notebook": "",
            "environment_variables": {},
            "cloudstorage": [],
        }

        server = UserServer(**{**base_parameters, **parameters})
        server.image_workdir = ""

        manifest = server._get_session_manifest()
        server_from_manifest = UserServer.from_js(user, manifest)

    assert expected in str(manifest)
    assert len(server_from_manifest.environment_variables) > 0
```

olevski (dismissed):
Thanks. Looks good!

olevski (approved):

PR #1168, opened 2022-07-15
fix(app): listing autosaves and debug statements

olevski:
Fixes minor issues with debug statements (some of which are useless but occur way to frequently).

And also with gitlab pagination when there are a lot of items to be listed. If the `iterator` option is used when listing things then the api handles the pagination seamlessly in the generator.

Depends on #1146


Panaetius (approved):
Thank you!

PR #1169, opened 2022-07-17
chore(app): fix for updating gitlab sdk to 3.6.0

olevski:
The newest version of the sdk initializes `client.user` to None even if `auth()` is not called. Older versions just do not set this at all if `auth()` is not called.

After this is merged then https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1146 and https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1168 can be merged.


mohammad-sdsc (approved):
👍

PR #1184, opened 2022-07-22
chore(app): document max lines for getting logs

olevski:
This uses the same webargs library for validation similar to other endpoints on the server.

Also I added the docs for this so that it shows up in swagger.


Panaetius (approved):
Thank you!

PR #1186, opened 2022-07-25
feat(app): use dataconf and telepresence

olevski:
This addresses two things:
- using telepresence for easier development
- using dataconf for managing the configuration

These are done together because the biggest problem in using telepresence was how to easily transfer or define the configuration used by the notebook service when running the notebook service locally.

With dataconf the notebook service uses the configuration from the following places:
- environment variables (highest precedence)
- config file specified in the `CONFIG_FILE` environment variable
- defaults from `renku_notebooks.config.default_config` (lowest precedence)

Now when someone wants to run the notebook service locally (with or without telepresence) they can simply make a copy of `example.config.hocon`, call it `.config.hocon`, fill in the missing values and run the notebook service. This does not replicate the environment from the cluster but it allows one to easily change the config and test things out. It also runs telepresence without mounting any drives from k8s into the local environment which has usually proved error prone.

/deploy #persist


olevski:
In addition to the above. Now also the notebook service does not have to be aware (or check) whether it is running with or without telepresence. When the environment from the deployment and storage was mounted from the cluster to the local environment then the notebook service had to check if telepresence is used or not and adapt.

lorenzo-cavazzi (changes requested):
The cleanup using the config file looks great!
And telepresence works well.

As discussed offline, choosing a different namespace might be easier. Would it make sense, for the time being, to slightly change the readme to mention that the namespace should be set in the kube config file under `context.namespace`? And perhaps also removing the possibility to pick a different namespace when running the `run-telepresence` script

olevski:
@lorenzo-cavazzi now you can specify `k8s {namespace = "test"}` in the `.config.hocon` file. That will make the notebook service run in the specified namespace.

lorenzo-cavazzi (approved):
The last solution works great!
Thanks for spending more time on this. Using telepresence on renku-notebooks is now really straightforward :rocket:

PR #1198, opened 2022-08-05
fix(app): catch upper/lowercase out of space error

olevski:
minor fix


olevski:
The error message is not always all lowercase. It can appear like this: `cannot create directory at 'ch04_data_wrangling/data_prep': No space left on device`. This should catch it regardless of the case.

olevski:
This problem show up in commit 132634d3 for https://renkulab.io/projects/the-graph-courses/intro-to-data-analysis-with-r-staging

lorenzo-cavazzi (approved):

PR #1199, opened 2022-08-08
chore: use 'sub' JWT field for user id

olevski:
This adds an `id` field to the user class.

The field is the keycloak user ID for registered user and for anonymous users it is the unique id that the gateway generates.

These are added in the annotations so that they can be used in publishing metrics. Currently the only user-identifying property we have is the email. But the id is better to use in logging or metrics because it better protects the user's privacy.

Also the `keycloak_user_id` field is removed because it is not used anywhere and anonymous users are not in keycloak.


Panaetius (approved):
thank you!

PR #1202, opened 2022-08-09
chore(docs): add roadmap

olevski:
I think this would be useful to guide and prioritize larger new features that should be added to the notebook service.

Let me know what you think. What should be added or removed? Should things be prioritized differently.

I have tried to just focus on bigger new features rather than fixes, maintenance and similar smaller features that extend existing functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,41 @@ Then it will redirect all traffik for the notebooks service from the deployment 
 port `8000`. Combininig telepresence with the steps above can be used to quickly test a notebook
 service in a full Renku deployment.
 
+## Roadmap
+
+This is just a quick overview for significant new feautures that will be added to the 
+notebook service. 
+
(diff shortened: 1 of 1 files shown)


olevski:
It is easiest if you look at the rendered reamde file here: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/chore-add-roadmap/README.md

olevski:
p.s. I converted the Gant chart to a normal list. It was pretty but not very useful.

olevski:
I will most likely close this PR and move these into a separate page in the `renku-python` github project where we have the Backlog, sprint, etc.

olevski:
Add a page in our github backlog/sprint planning project see https://github.com/orgs/SwissDataScienceCenter/projects/45/views/7

PR #1213, opened 2022-08-12
fix(app): usernames starting with numbers/symbols

### LINKED ISSUE #1157: Usernames that start with digits cause trouble in k8s
```"140XXXX\\\\\\\\\\\\\\\\\\\\\": a DNS-1035 label must consist of lower case alphanumeric characters or \\\\\\\\\\'-\\\\\\\\\\', start with an alphabetic character, and end with an alphanumeric character (e.g. \\\\\\\\\\'my-name\\\\\\\\\\',  or \\\\\\\\\\'abc-123\\\\\\\\\\', regex used for validation is \\\\\\\\\\'[a-z]([-a-z0-9]*[a-z0-9])?```

K8s services have to start with alphabetic characters. So usernames that start with a digit cannot start sessions.

### PR DESCRIPTION
**Pull request by olevski**
closes #1157 

/deploy #persist

### CODE REVIEW

File: renku_notebooks/util/kubernetes_.py [hunk 1]
```diff
@@ -56,11 +57,30 @@ def secret_exists(name, k8s_client, k8s_namespace):
...
+) -> str:
+    """Form a unique server name.
+
+    This is used in naming all of the k8s resources created by amalthea.
+    """
+    server_string_for_hashing = (
+        f"{safe_username}-{namespace}-{project}-{branch}-{commit_sha}"
+    )
+    safe_username_lowercase = safe_username.lower()
+    if re.match(r"[a-z]", safe_username_lowercase[0]):
```
Panaetius, line 71:
```suggestion
    if safe_username_lowercase[0].isalpha() and safe_username_lowercase[0].isascii():
```
No need for regex :slightly_smiling_face: Also then you can just lowercase on the `username=...` line since this works for both upper/lower case.

You could also just set prefix and overwrite in in an `if` then no `else` is needed.

Panaetius (changes requested):

Panaetius (approved):

olevski:
Sorry @Panaetius can you re-approve? The acceptance tests caught a problem I would have introduced with this PR because I accidentally removed the part that truncates the username. K8s names cannot be longer than 63 characters. That truncation was already in place I just accidentally removed it earlier. Now it is back.

Panaetius (approved):

PR #1219, opened 2022-08-16
feat(sidecar): run renku cli commands in rpc server

### LINKED ISSUE #1175: Add a save function to the session sidecar
This can be used by the ui to save unsaved work in a user's session when a user tries to shut down a session that has unsaved work.

olevski:
closes #1175 

Should be reviewed and merged only after #762 

/deploy #persist


Panaetius (commented):
the way it is implemented here, you wouldn't see valid parameters for an individual command (e.g. `renku save`) in the openapi specification, right? Maybe we could be a bit more explicit on which commands can be called and what parameters they have?.

Panaetius:
ah I assumed the code like `renku_run.__doc__ = ...` was picked up by openapi or similar

olevski:
> ah I assumed the code like renku_run.__doc__ = ... was picked up by openapi or similar

Yeah it was picked up by the json-rpc python library which makes a htlm page out of the methods with the `__doc__` sections. But I removed that part. Now I have a separate module where I set up all the rpc methods and their docs. So I do not have to do `renku_run.__doc__ = "..."`.

git_services/tests/conftest.py
```diff
@@ -36,18 +37,28 @@ def _commit_everything():
...
+        if init_renku:
+            check_call(
+                args=[
+                    "renku",
+                    "init",
+                    "--template-id=minimal",
+                    ".",
+                ],
+                cwd=git_cli.repo_directory,
+            )
```
Panaetius, line 61:
```suggestion
            check_call(
                args=[
                    "renku",
                    "init",
                    "--template-id=minimal",
                    git_cli.repo_directory,
                ],
            )
```
since path is an argument to init already.

git_services/git_services/sidecar/rpc_methods.py
```diff
@@ -0,0 +1,108 @@
+"""RPC Server methods.
+
+Note that all __doc__ of functions below are templated into a <pre></pre> section
```
Panaetius, line 3: There's no `__doc__` below.

Panaetius (changes requested):

git_services/git_services/sidecar/rpc_methods.py
olevski: Oh my bad that is leftover from when there was __doc__.

git_services/tests/conftest.py
olevski: Done.

Panaetius (approved):

PR #1223, opened 2022-08-22
feat(sidecar): git pull and discard unsaved changes

### LINKED ISSUE #1176: Add a "discard unsaved changes" function to the session sidecar
The ui can call this if a user has unsaved work in their session but wishes to not save it and discard it when they are shutting down their session.

olevski:
closes #1176 

Should be reviewed and merged only after #1219 and #762 are both merged.

/deploy #persist


Panaetius (approved):
thank you!

PR #1226, opened 2022-08-23
feat(session): check LFS size when cloning

### LINKED ISSUE #1144: Examine using `git lfs ls-files` to see if you have enough disk space when cloning
Potential problems:
- you still may run out of disk space because large files that should be in LFS are not in LFS
- you will have to parse the pretty-printed output from `git lfs ls-files` to determine the total size

Correction: `git lfs` version `3.2.0` should have json output: https://github.com/git-lfs/git-lfs/pull/5007

But you still have to deal with non-json output from "legacy" sessions that are not using the latest LFS version

olevski:
closes #1144 

This is a much more limited solution than what the issue was asking for. Namely the limitations are as follows:
- There is no easy way to check the size of the files that are outside of LFS, so if the files outside of LFS are larger than the free disk space there is no "advanced" error
- The message about how much more space is required cannot be easily "passed" from the init container to the notebook service, so the same solution as what we currently have is in place here - namely different errors have different exit codes. Then the notebook service can cross-reference the exit code from the init container and see that the problem was not enough disk space. It is not easy or clear how to pass through additional information with this system.

The main advantage of this is that the cloning step can use the LFS size and raise an exception rather than to wait for the disk to fill up and then capture the exception. Waiting on the disk to fill up and recognizing this case is not very reliable so doing this in a more reliable way is still better. Also with this change the user does not have to wait for the LFS files to be downloaded and then for the disk to run out of space. The LFS files do not need to be downloaded so that we can understand that we do not have enough space.

Lastly this check is only done when the user requests for the clone step to download the LFS data. If the clone step does not download any LFS data then the check is skipped.


Panaetius (dismissed):

olevski:
@Panaetius sorry for dismissing your approval.

There were conflicts I needed to resolve.

The confilcts were only in lines 135-138 in the new verison of git_services/git_services/init/cloner.py

olevski:
@Panaetius actually just hold of on this a bit. I will fix the tests and then ping you to approve.

Panaetius (approved):

PR #1230, opened 2022-08-25
feat(sidecar): symlink cloudstorage into project

### LINKED ISSUE #1165: Mount or link `/cloudstorage` folder in the user project directory
Users who use the s3 feature are sometimes confused about where they can access their data.

Therefore it would be nice to mount the s3 storage folder in the user's project directory.

Currently we use `/cloudstorage` to mount all the buckets in. And the users get the info about where to access the buckets when they specify the bucket credentials when they launch the sessions. But it is easy to miss this.

Panaetius:
closes #1165 

/deploy #persist extra-values=notebooks.cloudstorage.s3.enabled=true,notebooks.amalthea.extraChildResources[0].name=datasets,notebooks.amalthea.extraChildResources[0].group=com.ie.ibm.hpsys

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -28,6 +28,7 @@ jobs:
       renku-ui: ${{ steps.deploy-comment.outputs.renku-ui}}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled}}
       persist: ${{ steps.deploy-comment.outputs.persist}}
+      extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
     steps:
       - id: deploy-comment
         uses: SwissDataScienceCenter/renku-actions/check-pr-description@v1.1.0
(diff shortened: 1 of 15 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       renku-ui: ${{ steps.deploy-comment.outputs.renku-ui}}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled}}
       persist: ${{ steps.deploy-comment.outputs.persist}}
+      extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
```
olevski, line 31: Thanks a lot for spotting and adding this.

olevski (approved):
Thanks! Looks great!

olevski (approved):

olevski:
Thanks for incrementing the timeout for the acceptance tests @Panaetius! 

As I mentioned previously I am pretty sure this is the result of the sessions taking longer to startup. And the acceptance tests start and stop sessions many times so the effect is magnified.

Nevertheless I will open an issue to look into this a bit and see if we can alleviate it.

This PR is good to go. I will merge it.

Issue #1152, opened 2022-07-06
Add the session URL as an environment variable in all sessions


olevski:
This is useful for sessions that use dashboards through the jupyter-server-proxy. The proxy needs to know the path where the session runs so that it can properly modify the path used for any dashboards or similar apps that it routes.

olevski:
See what I had to do get dash plotly to run here: https://renkulab.io/projects/tasko.olevski/dash-jupyter-server-proxy-example

If I could get the session url from an environment variable then setting up dash would have been much easier

PR #1231, opened 2022-08-25
feat(app): add session url env var

mohammad-sdsc:
Adds a `SESSION_URL` env var that shows full session's url.

Fixes: #1152 

/deploy #persist


mohammad-sdsc:
Screenshot 2022-08-26 at 16 44 27

olevski (approved):
Thanks!

PR #1239, opened 2022-09-01
chore(app): reduce image size

olevski:
/deploy #persist

This is ready for review now. It adds a few changes:
- update the setup-python action to the latest version
- add caching for poetry in the actions that use it for faster builds (the new version of the python actions support poetry caching)
- switch the images to use alpine (reduces image size significantly)
- change images to use multistage builds and to exclude poetry from the images (i.e. poetry is used to generate the venv but is not really needed to run things at all) - this saves about 60MB per image
- remove resource limits (which throttle cpu) for the git sidecar and git clone container in a session, this was also contributed to sessions taking a long time to start 
- changed back the acceptance test timeout from 120min to 80 since the tests do not take that much time now

The only drawback is that the alpine based images take a while to build. More so than the debian based ones. But this can be resolved by adding some caching for the docker builds. I will resolve this in a subsequent PR.

For example the new alpine-based sidecar server images are now 84MB, compared to 240MB (with renku cli and with debian) and 124 (without renku cli but with debian).


Dockerfile
```diff
@@ -1,22 +1,24 @@
...
-    useradd -u 1000 -g kyaku -m kyaku
+FROM base as builder
+ENV POETRY_HOME=/opt/poetry
+ENV POETRY_VIRTUALENVS_IN_PROJECT=true
+ENV POETRY_VIRTUALENVS_OPTIONS_NO_PIP=true
+ENV POETRY_VIRTUALENVS_OPTIONS_NO_SETUPTOOLS=true
+COPY poetry.lock pyproject.toml ./
+RUN apk add --no-cache alpine-sdk libffi-dev && \
+    mkdir -p /opt/poetry && \
+    curl -sSL https://install.python-poetry.org | python3 - && \
```
Panaetius, line 14: I'm a bit worried that this could break things with suddenly changing the version out from under us. Should we pin this somehow?

olevski: Yeah good point. Pinned it to 1.2.1.

Panaetius (approved):

PR #1240, opened 2022-09-01
fix(sidecar): use fetch before running git status

olevski:
We should fetch before we run `git status` in the json rpc sidecar.


Panaetius (approved):
:rocket:

PR #1242, opened 2022-09-02
chore(ci): add docker layer caching

olevski:
🤞 I really hope this works because waiting for the exact same image to build 10 times in the same action is crazy.


Panaetius (approved):
Thank you, this is a really nice performance improvement

PR #1243, opened 2022-09-02
chore: track gh actions with dependabot

olevski:
According to the docs here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions

Dependabot can now keep actions up to date.


Panaetius (approved):
Please add to renku-python as well (though with target branch `develop`)

PR #1260, opened 2022-09-07
chore(chart): update datashim to 0.3.0

olevski:
/deploy #persist extra-values=notebooks.cloudstorage.s3.enabled=true,notebooks.amalthea.extraChildResources[0].name=datasets,notebooks.amalthea.extraChildResources[0].group=com.ie.ibm.hpsys

closes https://github.com/SwissDataScienceCenter/amalthea/issues/184


Panaetius (approved):

olevski:
After publishing the helm chart for datashim and deploying this on dev I can confirm that it works.

I will restart the acceptance tests so that they pass. Although the acceptance tests do not test s3 bucket mounting so it does not really matter but still.

PR #1293, opened 2022-09-28
fix(sidecar): renku needs to be on path

olevski:
There were some issues running renku cli commands from within the git rpc sidecar.

These were the result of:
- the python git library used by renku trying to use `bash` and the alpine images not having `bash` by default, after installing bash in the image (which added only 1MB to the size) I did not see this again
- the renku cli not being on the path - after I adjusted the path for the image everything works

cc @ciyer


Panaetius (approved):

ciyer (approved):
Thanks for this, this fixes the problems I was encountering.

PR #1294, opened 2022-09-28
fix: poetry flags not parsed properly

olevski:
Poetry 1.2.0 seems to have a bug where the settings set in environment variables are not parsed correctly and not really respected.

For example none of the images built are supposed to come with pip and setuptools. But both can be found in them.

If the settings are applied through `poetry config <config name> <value>` then pip and setuptools are truly excluded as originally intended. 

This has no real consequences other than keeping docker image size small.

/deploy #persist


Panaetius (approved):
thank you!

PR #1300, opened 2022-10-05
fix(sidecar): properly set and unset lfs auth

olevski:


olevski:
Not sure why acceptance tests do not trigger. Will open a new PR.

PR #1301, opened 2022-10-05
fix(sidecar): properly set and unset lfs auth

olevski:
/deploy #persist renku=fix-do-not-reload-page-after-failing-to-click-terminal

### DIFF
--- a/git_services/git_services/init/cloner.py
+++ b/git_services/git_services/init/cloner.py
@@ -1,15 +1,15 @@
-from contextlib import contextmanager
-import os
-import requests
-from datetime import datetime, timedelta
 import logging
-from time import sleep
-from pathlib import Path
+import os
 import re
-from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


git_services/git_services/init/cloner.py
```diff
@@ -90,19 +90,26 @@ def _setup_cloudstorage_symlink(self, mount_folder):
 
     @contextmanager
     def _temp_plaintext_credentials(self):
+        lfs_auth_setting = urljoin(self.git_url + "/", "info/lfs") + ".access"
```
Panaetius, line 93:
```suggestion
        lfs_auth_setting = urljoin(f"{self.git_url}/", "info/lfs.access")
```

olevski: @Panaetius I had to add a few more things in the setting name but I incorporated these changes.

Panaetius (approved):

PR #1310, opened 2022-10-13
fix(app): add jupyter server cache

olevski:
K8s does not cache CRDs at all (or very little at least). So listing servers or filtering them puts a lot of stress on the k8s api and results in really long response times.

Watching the events for jupyterservers and maintaining a cache that is used for this purpose will resolve this issue.

Thanks to @seanrmurphy for the fix.


olevski:
replaced by https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1323

PR #1323, opened 2022-10-27
feat(app): add jupyter server caching and multiple namespace support

olevski:
This is almost done needs a few final touches.

/deploy #persist


rokroskar:
This is looking pretty cool, thanks @olevski and @seanrmurphy! One stupid comment though: every time I see "js" I think JavaScript. I also expect that we will have other types of session servers shortly. Can we come up with a more generic name now?

olevski:
> Can we come up with a more generic name now?

Sure I can just call it "server".

js-watcher/cache_collection.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"context"
+	"log"
+
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type CacheCollection map[string]*Cache 
+
+func (c *CacheCollection) synchronize(ctx context.Context) {
```
seanrmurphy, line 12: I'm not sure why but a pointer receiver to a collection is not recommended - see https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

js-watcher/cache_collection.go
```diff
@@ -0,0 +1,103 @@
...
+	doneCh := make(chan bool)
+	for namespace, cache := range *c {
+		log.Printf("Starting sync for %s\n", namespace)
+		go cache.synchronize(ctx, doneCh)
+	}
+	syncCount := 0
+	log.Println("Waiting on cache synchronization")
+	for {
```
seanrmurphy, line 20: What about adding a timeout mechanism here in case of issues?

js-watcher/cache_collection.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return output, nil
+}
+
+func (c *CacheCollection) getByName(name string) ([]runtime.Object, error) {
+	output := []runtime.Object{}
+	for _, cache := range *c {
+		res, err := cache.getByName(name)
+		if err != nil {
+			continue
```
seanrmurphy, line 84: This looks v strange; usually if `err != nil` we need to handle it.

js-watcher/server.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+type Server struct {
+	caches CacheCollection
+	config Config
+	router *httprouter.Router
+	server *http.Server
```
seanrmurphy, line 16: A server inside a server looks v strange to me; also, having the handler as some kind of parent struct really give off some kind of code smell to me; however, I don't know exactly what is the better approach rn...

js-watcher/server.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type LoggingResponseWriter struct {
+	http.ResponseWriter
+	req http.Request
+}
+
+func (r LoggingResponseWriter) WriteHeader(status int) {
```
seanrmurphy, line 24: Did you think of using a middleware for this such as https://echo.labstack.com/middleware/

js-watcher/routes.go
```diff
@@ -0,0 +1,53 @@
+package main
```
seanrmurphy, line 1: I really think it makes sense to have all functions operating on the same struct in a single file - could we merge them?

js-watcher/cache.go
```diff
@@ -0,0 +1,135 @@
...
+	k8sCache "k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+type Cache struct {
+	lister    k8sCache.GenericLister
+	informer  k8sCache.SharedIndexInformer
+	namespace string
+	userIdLabel string
```
seanrmurphy, line 29: not `gofmt` compliant - at least run it through that. I'd also recommend addressing `revive` comments generally, perhaps not the comment ones but at least the ones on variable and function names....

js-watcher/cache.go
```diff
@@ -0,0 +1,135 @@
...
+	return output, nil
+}
+
+func NewCacheFromConfig(ctx context.Context, config Config, namespace string) (*Cache, error) {
+	var err error
+	var clientConfig *rest.Config
+	clientConfig, err = rest.InClusterConfig()
+	if err != nil {
+		log.Println("Cannot setup in-cluster config, looking for kubeconfig file")
+		var kubeconfig *string
```
seanrmurphy, line 110: Could we break out this `kubeconfig` sutff into a separate function - it kinda distracted me from the main point of this function...

js-watcher/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.19.0-alpine3.16 as builder
+COPY . /src
```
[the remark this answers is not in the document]
olevski: I would expect that the Dockerfile uses the current code to build my image. It should not clone. 

We get reproducibility by pinning requirements and pinning our built images in our CI pipelines. But cloning here would just make development faster. Chartpress takes care of publishing the image with the right tag.

js-watcher/main.go
```diff
@@ -0,0 +1,46 @@
+package main
+
+import (
+	"context"
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/julienschmidt/httprouter"
```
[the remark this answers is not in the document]
olevski: Gin uses a custom version of this, see [here](https://gin-gonic.com/docs/benchmarks/). Also the library itself [says](https://github.com/julienschmidt/httprouter#web-frameworks-based-on-httprouter) that a bunch of other libraries that provide fancier functionality use it as a dependency. `julienschmidt/httprouter` has zero dependencies.

We do not need super-fancy route handling so I thought we can simply get away with using smaller requirements.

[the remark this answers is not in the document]
olevski: I did not want to use gorillamux because they do not have a maintainer and according to the [readme](https://github.com/gorilla/mux#gorillamux) on their repo are still looking for one.

js-watcher/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+func main() {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+  	stopCh := make(chan os.Signal, 1)
+	signal.Notify(stopCh, syscall.SIGTERM, os.Interrupt)
+	
+	config := NewConfigFromEnvOrDie("JS_CACHE_")
+	log.Printf("Running with config %+v\n", config)
```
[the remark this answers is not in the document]
olevski: No, nothing is secret in the config. That is why I print the whole thing when the server starts.

js-watcher/main.go
```diff
@@ -0,0 +1,46 @@
...
+		log.Println("Received shutdown signal")
+		log.Println("Shutting down server gracefully")
+		err := aServer.server.Shutdown(ctx)
+		if err != nil {
+			log.Printf("Graceful server shutdown failed: %v\n", err)
+		}
+		cancel()
+	}()
+	go aServer.caches.run(ctx)
+	aServer.caches.synchronize(ctx)
```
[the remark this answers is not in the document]
olevski: The whole server will block until the cache fully synchronizes. That is why that is not a go routine. Once the cache is synced then all resources should be in memory and the watch should handle updating things.

Doing `run` just starts the watches the informer uses internally and it kicks off the informer's cache sync. Only when the synchronize step completes then the cache is ready and the server can respond to any requests.

js-watcher/main.go
```diff
@@ -0,0 +1,46 @@
...
+	log.Println("Setting up routes")
+	aServer.routes()
```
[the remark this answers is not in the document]
olevski: Will rename to `registerRoutes`

js-watcher/cache.go
```diff
@@ -0,0 +1,135 @@
...
+
+type Cache struct {
+	lister    k8sCache.GenericLister
+	informer  k8sCache.SharedIndexInformer
+	namespace string
+	userIdLabel string
+}
+
+type GenericResource struct {
+	metav1.TypeMeta   `json:",inline"`
```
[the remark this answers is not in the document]
olevski: Ok I will move it last and add a line break.

js-watcher/cache.go
```diff
@@ -0,0 +1,135 @@
...
+	doneCh <- true
+}
+
+// Run the informer.
+func (c *Cache) run(ctx context.Context) {
+	c.informer.Run(ctx.Done())
+}
+
+func (c *Cache) getAll() ([]runtime.Object, error) {
+	output, err := c.lister.List(labels.NewSelector())
```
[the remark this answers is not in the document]
olevski: Sure I can rename to `res`. No problem.

js-watcher/cache_collection.go
olevski: Np I will change it, I initially was scared of using a value receiver here because of potentially copying the caches. But the map has pointers to the caches as elements - so it is all good. The pointer addresses to the caches will be copied not the caches themselves.

js-watcher/main.go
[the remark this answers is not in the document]
seanrmurphy: Thanks - just wanted to confirm...

js-watcher/cache_collection.go
olevski: 👍 but I will just do `log.Fatal` on a timeout there. Or at most retry a few times and then eventually crash the server. There is no point in running the server if we cannot be confident that step completed successfully.

js-watcher/main.go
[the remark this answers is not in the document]
seanrmurphy: I think I find the logic here somewhat unclear and I think the abstraction could be improved. What you think of something like `server.initialize()` which contains the `caches.run()`, `caches.synchronize()`, `routes()` and `setup()` calls followed by `server.start()` - that way it's at least possible in principle to track the `caches.run()` goroutines inside the `Server` object if required...

js-watcher/cache.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (c *Cache) getByNameAndUserId(name string, userId string) (runtime.Object, error) {
+	output, err := c.getByName(name)
+	if err != nil {
+		return nil, err
+	}
+	outputUnstructured, err := runtime.DefaultUnstructuredConverter.ToUnstructured(output)
+	if err != nil {
+		return nil, fmt.Errorf("could not convert server %s to unstructured: %w", name, err)
```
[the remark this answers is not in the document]
seanrmurphy: Cool - thanks!

js-watcher/cache_collection.go
seanrmurphy: Whoops - I meant to a `map` obviously...

js-watcher/cache_collection.go
seanrmurphy: Sure - it's more about catching the case where it ends up waiting (for whatever reason) and never errors and never enters operation...

js-watcher/server.go
olevski: Yeah if we switch to a more extensive server library that is easy. But you can make middlewares even with the built-in http server library. The problem is that middlewares for `net/http` take a http [HandlerFunc](https://pkg.go.dev/net/http#HandlerFunc) and return another one. But in the http HandlerFunc we only get a [`ResponseWriter` interface](https://pkg.go.dev/net/http#ResponseWriter) and this does not allow me to read the response or status code. Only write them (in some unreachable location). That is why I had to wrap the response writer like this so that I can access the status code.

olevski: For now I dont want to re-write this whole thing with echo or gin just for the sake of better logging. I can open a PR for that as an improvement down the road.

js-watcher/cache_collection.go
olevski: I never return 404 on looking for a resource that does not exist. I simply return an empty list if you specify a server name that is not currently present. Luckily `k8s.io/apimachinery/pkg/api/errors` provides `IsNotFound` which checks if a k8s error is triggered by the resource missing. So I will `continue` only in this case.

js-watcher/main.go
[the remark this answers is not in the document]
olevski: Yeah that makes sense.

rokroskar:
> Sure I can just call it "server".

Then we call it the "ss-watcher"? Sounds... maybe problematic ;)

olevski:
@rokroskar how about calling it `k8s-watcher`. Because it will watch any resource you mention in the config. Not just servers.

rokroskar:
Perfect!

js-watcher/routes.go
olevski: I really want someone to take one quick look and see all the http routes that are exposed. Combining things in one file and removing `routes.go` file makes this harder.

I renamed `routes.go` to `server_routes.go` to indicate they are related.

js-watcher/cache.go
olevski: done

js-watcher/cache.go
olevski: ran it through gofmt and revive and cleared all errors

js-watcher/server.go
olevski: Embedding `htts.Server` as I suggested above works. And it removes all cases of `server.server` weirdness except when we construct a new server. But I think this is better than what I had previously.

js-watcher/cache_collection.go
olevski: Added a timeout. It is configurable with a default value of 10 mins for all caches to sync.

js-watcher/cache.go
```diff
@@ -0,0 +1,150 @@
...
+	return res, nil
+}
+
+// getByName retrieves a specific resource from the informer cache.
+func (c *Cache) getByName(name string) (res runtime.Object, err error) {
+	res, err = c.lister.Get(fmt.Sprintf("%s/%s", c.namespace, name))
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
```
seanrmurphy, line 90: Not sure if you're deliberately trying to be explicit here but the return vals do not need to be specified here; in fact, I think you can leave out the `return` completely.

js-watcher/cache.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse metadata on server %s: %w", name, err)
+	}
+	if resTyped.Metadata.Labels[c.userIDLabel] != userID {
+		return nil, nil
+	}
+	return res, nil
+}
+
+func initializeK8sDynamicClient() (ks8DynamicClient dynamic.Interface, err error) {
```
seanrmurphy, line 115: ks8DynamicClient -> k8sDynamicClient

js-watcher/cache.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Cache is a light wrapper around a k8s informer for a single k8s namespace.
+type Cache struct {
+	lister      k8sCache.GenericLister
+	informer    k8sCache.SharedInformer
+	namespace   string
+	userIDLabel string
+}
+
+// GenericResource allows unmarshalling the metadata of dynamic resources without
```
seanrmurphy, line 33: GenericResource -> GenericKubernetesResource

js-watcher/server_routes.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// routers registers the handlers for all http endpoints the server supports.
+func (s *Server) registerRoutes() {
+	s.router.HandlerFunc("GET", "/servers", s.handleIndex())
```
seanrmurphy, line 11: I think you can do this which is prob neater:

```
func (s *Server) registerRoutes() {
	s.router.HandlerFunc("GET", "/servers", s.handleIndex)
}

func (s *Server) handleIndex(w http.ResponseWriter, req *http.Request) {
		output, err := s.caches.getAll()
		s.respond(w, req, output, err)
}
```

seanrmurphy (commented):
Just a couple more comments...

js-watcher/cache.go
olevski: Oh cool I did not know that. It cannot be used in all cases. But in most I can and I did so.

It feels wrong coming from python. Because that always results in None being returned in python.

olevski:
I am pretty sure the integration tests are failing because there are not enough resources in the github runners to schedule all the loads. We were really cutting it close and now with the addition of the k8s resource caching we are over the limit. By default you really do not get a lot for CI jobs. So I will run the integration tests locally and double check.

Panaetius (changes requested):

Panaetius (approved):

PR #1329, opened 2022-11-10
feat: run sessions in subdomain

### LINKED ISSUE #1322: Run each session with its own subdomain
Currently sessions run as a path on `renku.domain.com/sessions/<session-id>`. The session id is simply the name of the jupyterServer in k8s.

What we want to have is running each session in its own subdomain. So something like: `<session-id>.sessions.renku.domain.com`

This has a few benefits:
- security
- we would be able to ssh into sessions
- we would be able to connect vs code to a running session (this uses ssh under the hood)

We need to check for several things:
- cors and what setup is required to make cors work
- limits on tls certificates for lets encrypt (it would be nice if we can have one tls cert for all session subdomains)
- keycloak setup and valid callbacks for authenticating access to sessions for authenticated users
- if we have to have one tls cert per session how does this affect letsencrypt quotas and how do those apply (there are different rules for subdomains, etc)

Panaetius:
closes #1322 

/deploy #persist renku=keycloak-18.3 amalthea=1322-session-subdomains extra-values=notebooks.sessionIngress.tlsSecret=dev-renku-ch-tls-wildcard,notebooks.sessionIngress.host=dev.renku.ch

### DIFF
--- a/helm-chart/renku-notebooks/templates/statefulset.yaml
+++ b/helm-chart/renku-notebooks/templates/statefulset.yaml
@@ -67,6 +67,12 @@ spec:
               value: {{ .Values.sessionIngress.tlsSecret }}
             - name: NB_SESSIONS__INGRESS__ANNOTATIONS
               value: |
+                nginx.ingress.kubernetes.io/configuration-snippet: |
+                  more_set_headers "Content-Security-Policy: frame-ancestors 'self' https://*.{{ .Values.sessionIngress.host }}";
+                nginx.ingress.kubernetes.io/enable-cors: "true"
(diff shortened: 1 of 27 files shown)


Panaetius:
Needs Valid callback URL on keycloak jupyter-server client set to '*' for now to work

olevski:
@Panaetius this can be closed, correct? Please close it if you do not need it.

Panaetius:
Well it's not needed yet but I'm not sure if we still want this at some point in the future? But it's already outdated so maybe not worth keeping it around

PR #1336, opened 2022-11-25
fix: accept zero as max_lines parameter for logs

olevski:
/deploy #persist

### DIFF
--- a/renku_notebooks/api/notebooks.py
+++ b/renku_notebooks/api/notebooks.py
@@ -296,7 +296,7 @@ def server_options(user):
     {
         "max_lines": fields.Integer(
             load_default=250,
-            validate=validate.Range(min=1, max=None, min_inclusive=True),
+            validate=validate.Range(min=0, max=None, min_inclusive=True),
         )
     },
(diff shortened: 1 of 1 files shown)


ciyer:
This change makes sense, but it does't fix the problem that we were having. The backend still responds with this error when we try to make a query with `max_lines=0`:

```
{"error": {"code": 1422, "message": "The request was well-formed but was unable to be followed due to semantic errors.", "detail": null}}
```

olevski:
> This change makes sense, but it does't fix the problem that we were having. The backend still responds with this error when we try to make a query with `max_lines=0`:
> 
> ```
> {"error": {"code": 1422, "message": "The request was well-formed but was unable to be followed due to semantic errors.", "detail": null}}
> ```

@ciyer there were actually 2 places where this needs to be changed. The first one I chaged is only for documentation purposes in the swagger page. Now it should truly work.

ciyer:
Yes, this fixes the problem, thanks!

ciyer (approved):
👍

PR #1338, opened 2022-11-30
fix: handle Gitlab token expiry

olevski:
/deploy #persist renku-gateway=master renku=master

This (unlike #1335) receives the refresh token in each session. Then the renku refresh token is used to keep the access token continuously valid and to exchange it for a valid gitlab token when needed.

closes https://github.com/SwissDataScienceCenter/renku-gateway/issues/606


renku_notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -47,7 +47,37 @@ def main(server: "UserServer"):
...
+                                "name": "RENKU_ACCESS_TOKEN",
+                                "value": str(server._user.access_token),
+                            },
+                            {
+                                "name": "RENKU_REFRESH_TOKEN",
+                                "value": str(server._user.refresh_token),
+                            },
+                            {
+                                "name": "RENKU_REALM",
+                                "value": "Renku",
```
mohammad-alisafaee, line 66: I couldn't find anywhere in `renku` repo's values files if this is configurable or not. If it is, then it might be difficult to find this instance here upon a change in the config value (although highly improbable). I'm not sure how easy it is to use a configurable value here.

mohammad-alisafaee (changes requested):
Thanks Tasko! This looks great! Please see my comments.

renku_notebooks/api/amalthea_patches/git_proxy.py
olevski: So this is hardcoded already pretty much everywhere it is used. I remember checking if there is a value I could pick up in the values files and there wasn't. I will check again.

But if this is not in the values file (whcih I am 99% sure it is not) then I propose we leave this for a full helm chart refactoring. Rather than adding just this value somewhere in the values file just to have it removed/refactored down the line or to have it not-hardcoded here but hardcoded in all other places.

olevski: So yeah you are right. We do have an optional spot in the global values file for the name of the keycloak realm. I switched to using that if it is defined and defaulting to "Renku" if it is not provided.

mohammad-alisafaee (dismissed):

mohammad-alisafaee (approved):
🚀

PR #1374, opened 2022-12-30
feat(app): experimental azure blob storage support

### LINKED ISSUE #1354: support azure blob storage on the azure cloud PoC
**Disclaimer**: This is not about properly support Azure Blob storage, only for enabling it for a PoC to test it.

Add support for mounting Azure Blob storage in Sessions on the Azure PoC. We can repurpose the S3 fields in the UI and have a git branch of the notebooks service that uses them to mount blob storage instead of S3.

We probably need to install a custom CSI driver instead of datashim for this to work.

olevski:
closes #1354 

This can be merged and released. Azure blob support is disabled by default. We can simply enable it where we want it as long as the cluster already has the azure blob csi driver.

There is also some very light refactoring for the following:
- Handle mounting more than one cloud storage type.
- Use different classes for Jupyter servers on creation requests and on listing requests. This is useful because when creating a server we need to a bunch of checks that we do not really care about when simply listing an existing server.

/deploy #persist


renku_notebooks/api/notebooks.py
```diff
@@ -94,13 +99,19 @@ def user_servers(user, **query_params):
         - servers
     """
     servers = [
-        UserServer.from_js(user, js, config.k8s.client)
-        for js in config.k8s.client.list_servers(user.safe_username)
+        UserServerManifest(i)
+        for i in config.k8s.client.list_servers(user.safe_username)
```
Panaetius, line 103: `i` is convention for loop index (int), we should use something else since it's a server (`s` at the least)

renku_notebooks/api/classes/cloud_storage/existing.py
```diff
@@ -0,0 +1,55 @@
...
+                    and patch["value"].get("kind") == "PersistentVolume"
+                    and patch["value"].get("spec", {}).get("csi", {}).get("driver")
+                    == "blob.csi.azure.com"
+                    and endpoint_annotation
+                    in patch["value"].get("metadata", {}).get("annotations", {})
+                    and patch["value"]
+                    .get("spec", {})
+                    .get("csi", {})
+                    .get("volumeAttributes", {})
+                    .get("containerName")
```
Panaetius, line 33: Would be nice if this if could be broken down to something more readable, like
```
is_persistent_volume = patch["value"].get("kind") == "PersistentVolume"
driver_is_azure = patch["value"].get("spec", {}).get("csi", {}).get("driver")== "blob.csi.azure.com"
has_container_name = patch["value"]
                    .get("spec", {})
                    .get("csi", {})
                    .get("volumeAttributes", {})
                    .get("containerName")

if isinstance(patch["value"], dict) and is_persistent_volume and driver_is_azure and has_container_name [...]:
```
then the if is a bit more readable. But not strictly necessary

renku_notebooks/api/schemas/cloud_storage.py
```diff
@@ -15,21 +17,43 @@ class Meta:
...
+            and config.cloud_storage.azure_blob.enabled
+        ):
+            cloud_storage = AzureBlobRequest(
+                endpoint=data["endpoint"],
+                container=data["bucket"],
+                credential=data["secret_key"],
+                mount_folder=config.cloud_storage.mount_folder,
+                read_only=config.cloud_storage.azure_blob.read_only,
+            )
+        else:
```
Panaetius, line 38: Shouldn't we check that s3 is enabled here via `config.cloud_storage.s3.enabled`?

Also we should fail is both s3.enabled=True and azure_blob.enabled=True in the values file somewhere (I didn't see code for this)

Panaetius (changes requested):

renku_notebooks/api/schemas/cloud_storage.py
olevski: @Panaetius for now there is no problem if both s3 and azure blob storage are enabled. And I foresee that some deployments will want to use both. Especially those running on Azure. But even non-Azure deployments can use both s3 and azure provided that they have installed the azure blob csi drivers in their k8s cluster.

The way I recognize azure blob storage is if the access key is not provided by the secret key is provided. That is the only way. It is not perfect but it should work to start. Then we can adapt the ui and have proper indication what storage type is being used. But for now we are jumping onto the same api for the s3 storage. 

When the user mounts a azure blob container then the endpoint is still the azure endpoint, the bucket is the azure blob container name and the secret key is the azure blob credentials. And the public key is left blank.

So I will add the checks for what is enabled and fail if s3 stuff is provided but s3 is disabled. But we can definitely support both azure blob and s3.

Panaetius (approved):

PR #1389, opened 2023-01-18
feat: ssh access to sessions

olevski:
Enables ssh access in notebooks via a jump host.


olevski:
> is this just temporary until we can expose the port properly?

Yes. And exposing the port will only for dev.renku.ch and production deployments. Not every single CI or personal deployment.

Panaetius (dismissed):

Panaetius (approved):

Issue #1311, opened 2022-10-13
Ensure injecting environment variables in sessions does not break sessions


olevski:
We need to make sure that we do not overwrite existing environment variables in sessions with the ability to inject them in sessions.

Currently we do not do any checks to see if existing variables are overwritten.

olevski:
Since the env variables are just patches on the amalthea manifest. We should also see what happens if you try to patch an existing variable. Or if there are two patches that assign different values to the same variable.

Ideally and most conservatively we simply error out whenever we see that an environment variable that already exists is being overwritten.

PR #1396, opened 2023-01-31
feat(session): fail when injecting env vars that exist in session

mohammad-alisafaee:
Checks that user doesn't override env vars sets by Amalthea patches and that Amalthea patches don't override the same env var with different values (in the same container).

Fixes: #1311

### DIFF
--- a/renku_notebooks/api/classes/server.py
+++ b/renku_notebooks/api/classes/server.py
@@ -16,7 +16,7 @@ from ..amalthea_patches import init_containers as init_containers_patches
 from ..amalthea_patches import inject_certificates as inject_certificates_patches
 from ..amalthea_patches import jupyter_server as jupyter_server_patches
 from ...config import config
-from ...errors.programming import ConfigurationError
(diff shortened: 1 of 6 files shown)


renku_notebooks/api/classes/server.py
```diff
@@ -364,6 +367,29 @@ def _get_session_manifest(self):
         }
         return manifest
 
+    def _check_environment_variables_overrides(self, patches_list):
```
olevski, line 370: I think this should be a static method since we do not need "self" at all.

renku_notebooks/api/notebooks.py
```diff
@@ -169,7 +169,7 @@ def launch_notebook(
     image,
     server_options,
     environment_variables,
-    cloudstorage=[],
+    cloudstorage=None,
```
olevski, line 172: Thanks for catching this and correcting it.

olevski (changes requested):
One small thing. All else looks good.

Thank you!

olevski (approved):

PR #1406, opened 2023-02-15
fix(sidecar): anon users clone repo at correct SHA

olevski:
If an anonymous user requests a session at a commit sha other than head then they would always get head.

This happened only for anonymous sessions. This fixes the problem.

/deploy #persist


olevski:
Thanks @lorenzo-cavazzi for reporting.

Panaetius (approved):

lorenzo-cavazzi (approved):
It works well with this fix :+1:

PR #1407, opened 2023-02-15
fix(app): show if ssh enabled in /version

olevski:
/deploy #persist


lorenzo-cavazzi (approved):
The fix works well :+1:

Panaetius (approved):

PR #1416, opened 2023-03-02
fix(git-proxy): refresh token expiry check

olevski:
/deploy #persist

Fixes this error:

```
  |   | 2023-03-02 16:16:01 | {"log":"\u0009/src/config/config.go:159 +0xaf4 ","stream":"stderr","time":"2023-03-02T15:16:01.317253742Z"}
  |   | 2023-03-02 16:16:01 | {"log":"created by github.com/SwissDataScienceCenter/renku-notebooks/git-https-proxy/config.ParseEnv ","stream":"stderr","time":"2023-03-02T15:16:01.317241681Z"}
  |   | 2023-03-02 16:16:01 | {"log":"\u0009/src/config/config.go:310 +0xb1 ","stream":"stderr","time":"2023-03-02T15:16:01.317229952Z"}
  |   | 2023-03-02 16:16:01 | {"log":"github.com/SwissDataScienceCenter/renku-notebooks/git-https-proxy/config.(*GitProxyConfig).periodicTokenRefresh(0xc00010add0) ","stream":"stderr","time":"2023-03-02T15:16:01.317217412Z"}
  |   | 2023-03-02 16:16:01 | {"log":"\u0009/src/config/config.go:297 +0x112 ","stream":"stderr","time":"2023-03-02T15:16:01.317204963Z"}
  |   | 2023-03-02 16:16:01 | {"log":"github.com/SwissDataScienceCenter/renku-notebooks/git-https-proxy/config.(*GitProxyConfig).isJWTExpired(0xc00010add0, {0xc000024014, 0x28b}) ","stream":"stderr","time":"2023-03-02T15:16:01.31719131Z"}
  |   | 2023-03-02 16:16:01 | {"log":"goroutine 4 [running]: ","stream":"stderr","time":"2023-03-02T15:16:01.31717943Z"}
  |   | 2023-03-02 16:16:01 | {"log":" ","stream":"stderr","time":"2023-03-02T15:16:01.31716735Z"}
  |   | 2023-03-02 16:16:01 | {"log":"[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6ae432] ","stream":"stderr","time":"2023-03-02T15:16:01.317151881Z"}
  |   | 2023-03-02 16:16:01 | {"log":"panic: runtime error: invalid memory address or nil pointer dereference ","stream":"stderr","time":"2023-03-02T15:16:01.317017242Z"}
```

The error resulted from swapping to the offline access refresh token that never expire. So the `exp` claim in the token is completely omitted. But in the code I was accessing it with the assumption that it is always there.

Now I use a method from the `jwt` library specifically for this case. And the library properly handles the case of a missing exp claim. Since the claim is omitted from tokens that do not expire - we assume that if the exp claim is missing the token does not expire.

### DIFF
--- a/git-https-proxy/config/config.go
+++ b/git-https-proxy/config/config.go
@@ -152,7 +152,7 @@ func ParseEnv() *GitProxyConfig {
 		SessionTerminationGracePeriod: SessionTerminationGracePeriod,
 		gitAccessTokenLock:            &sync.RWMutex{},
 		renkuAccessTokenLock:          &sync.RWMutex{},
-		expiredLeeway:                 time.Second * 30,
+		expiredLeeway:                 time.Second * time.Duration(refreshCheckPeriodSecondsParsed) * 4,
 		refreshTicker:                 time.NewTicker(time.Second * time.Duration(refreshCheckPeriodSecondsParsed)),
 	}
 	// Start a go routine to keep the refresh token valid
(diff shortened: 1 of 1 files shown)


olevski:
The change to the `expiredLeeway` is not the real fix. But it should be set as a multiple the period for how often we check if the token is expired. This gives us the chance to try to refresh the token more than once before it fully expires. And it ensures that we will not run ever with an expired token (we want to avoid this for sure).

I.e. if you check that the token expires every 30 minutes but you try to refresh 5 minutes before the token expires then you will have potentially some time (25 minutes in the worse case?) where the token is expired and no attempt has been made to refresh it.

Panaetius (approved):

PR #1420, opened 2023-03-10
chore: use server instead of notebook

rokroskar:
We can use the generic `jupyter server` to (hopefully) reliably set the various defaults. 

/deploy #persist


olevski:
Superseded by https://github.com/SwissDataScienceCenter/renkulab-docker/pull/320

PR #1472, opened 2023-04-28
fix: do not fail on submodule errors

### LINKED ISSUE #1466: Do not fail on git submodule init
Currently when the repo is being cloned if there are git submodules and they are not setup properly the cloning will fail.

This should not be the case because many times git submodules are very easy to setup in a half-working or fully broken way.

Panaetius:
closes #1466


olevski (approved):
Thanks @Panaetius sorry for letting this slip through the cracks.

PR #1483, opened 2023-05-15
feat(app): use crc service

olevski:
Uses the compute resource access control service for launching sessions.


Dockerfile
```diff
@@ -1,23 +1,29 @@
-FROM python:3.8-alpine as base
-RUN apk add --no-cache curl tini && \
-    adduser -u 1000 -g 1000 -D kyaku
-WORKDIR /home/kyaku/renku-notebooks
-
-FROM base as builder
-ENV POETRY_HOME=/opt/poetry
+FROM python:3.11-bullseye as builder
```
rokroskar, line 1: python 3.10 is used elsewhere - is that on purpose?

olevski: @rokroskar the other components install the renku cli. And I think the CLI is still not fully compatible with python 3.11. This one does not use the CLI so we can go to 3.11.

rokroskar: if you don't need 3.11 we could keep it consistent?

renku_notebooks/api/classes/crc.py
```diff
@@ -0,0 +1,147 @@
...
+        headers = None
+        if user.access_token is not None:
+            headers = {"Authorization": f"bearer {user.access_token}"}
+        res = requests.get(self.crc_url + "/resource_pools", headers=headers)
+        if res.status_code != 200:
+            raise IntermittentError(
+                message="The compute resource access control service sent "
+                "an unexpected response, please try again later",
+            )
+        resource_pools = res.json()
```
Panaetius, line 101: This code is duplicated three times (here and above), maybe it should be its own method?

renku_notebooks/api/classes/server.py
```diff
@@ -221,40 +222,21 @@ def _get_registry_secret(self, b64encode=True):
...
             "limits": {"memory": mem},
         }
         if config.sessions.enforce_cpu_limits == "lax":
-            if "cpu_request" in config.server_options.ui_choices:
-                resources["limits"]["cpu"] = max(
-                    config.server_options.ui_choices["cpu_request"]["options"]
-                )
-            else:
-                resources["limits"]["cpu"] = cpu_request
+            resources["limits"]["cpu"] = 3 * cpu_request
```
Panaetius, line 234: What's the `3` here?

renku_notebooks/api/schemas/server_options.py
```diff
@@ -1,82 +1,159 @@
...
+        )
+
+    @classmethod
+    def from_request(cls, data: Dict[str, Any]) -> "ServerOptions":
+        """Convert a server options request dictionary to the model."""
+        return ServerOptions(
+            cpu=data["cpu_request"],
+            gpu=data["gpu_request"],
+            memory=data["mem_request"],
+            default_url=data["defaultUrl"],
```
Panaetius, line 119: We have a weird mix of camel case and snake case in request fields, we should clean this up at some point

Panaetius (changes requested):
thank you! nice to see this taking shape :slightly_smiling_face:

Dockerfile
olevski: Ok cool I will update to 3.11.

Dockerfile
```diff
@@ -1,23 +1,29 @@
...
+RUN groupadd --gid 1000 renku && \
+    useradd --gid 1000 --uid 1000 --groups 100 --create-home renku && \
+    mkdir -p /app && \
+    chown -R 1000:1000 /app 
+USER 1000:1000
+WORKDIR /app
+RUN python3 -m pip install --user pipx && \
+    python3 -m pipx ensurepath && \
+    /home/renku/.local/bin/pipx install poetry && \
```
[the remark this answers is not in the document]
olevski: @Panaetius I use pipx because I want poetry to be in a separate environment and I do not want to have poetry at all in the final image.

So either I use pipx or I create a separate virtualenv for poetry and then install poetry there. Pipx was easier to use. So I will leave this as is.

renku_notebooks/api/schemas/server_options.py
olevski: I will open an issue for this. We should coordinate with the UI.

renku_notebooks/api/classes/server.py
olevski: This is to allow the cpu to spike up momentarily. Because if we set the limit on the cpu it gets throttled and then when you create a new session it takes a really long time for the session to start up. If the limits are set to "lax" (i.e. "relaxed") then the limit is larger than the request. But if the limits are set to strict then the limit is the same as the request and the cpu is throttled.

git_services/Dockerfile.sidecar
```diff
@@ -1,24 +1,30 @@
...
+FROM python:3.10-bullseye as builder
+RUN groupadd --gid 1000 renku && \
+    useradd --gid 1000 --uid 1000 --groups 100 --create-home jovyan && \
+    mkdir -p /app && \
+    chown -R 1000:1000 /app 
+USER 1000:1000
+WORKDIR /app
+RUN python3 -m pip install --user pipx && \
+    python3 -m pipx ensurepath && \
+    /home/jovyan/.local/bin/pipx install poetry && \
```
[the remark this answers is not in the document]
olevski: I will leave this as is as we discussed.

renku_notebooks/api/classes/crc.py
```diff
@@ -0,0 +1,147 @@
...
+            raise IntermittentError(
+                "The CRC sent an unexpected response, please try again later."
+            )
+        pools = res.json()
+        default_pools = [p for p in pools if p.get("default", False)]
+        if len(default_pools) < 1:
+            raise ConfigurationError("Cannot find the default resource pool.")
+        default_pool = default_pools[0]
+        default_classes = [
+            cls for cls in default_pool.get("classes", []) if cls.get("default", False)
```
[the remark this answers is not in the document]
olevski: image

The classes field is a list. Not a dictionary.

renku_notebooks/api/schemas/server_options.py
olevski: https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1525

Dockerfile
rokroskar: oh it does? then we can build 3.11 base images as well?

renku_notebooks/api/classes/crc.py
```diff
@@ -0,0 +1,147 @@
...
+        if storage < 1:
+            raise InvalidComputeResourceError(
+                message="Storage requests have to be greater than or equal to 1GB."
+            )
+        if storage > res_class.get("max_storage"):
+            raise InvalidComputeResourceError(
+                message="The requested storage surpasses the maximum value allowed."
+            )
+        # Memory and disk space in CRC are assumed to be in gigabytes whereas
+        # the notebook service assumes that if a plain number is used then it is bytes.
```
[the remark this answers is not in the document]
olevski: So the comment is not right. I updated the schema for the request in launch_notebook to be in gigabytes. But I did not update the comment. Now I added the assumption that disk and memory are in gigabytes in the docstring.

Dockerfile
olevski: The problem I had was that I could not get numpy installed on the debian images I was using with python 3.11. But now since we removed numpy from the dependencies things work.

olevski:
@Panaetius I addressed all the comments. The integration tests are failing because it seems that the oauth application used by the tests has been removed from gitlab. I will fix this later. I ran the integration tests locally and they all passed.

renku_notebooks/api/classes/crc.py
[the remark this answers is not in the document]
Panaetius: sorry my bad, I got mixed up with the `cls.get` after and thought it should be dict...

renku_notebooks/api/classes/server.py
Panaetius: makes sense. you should give it a descriptive name, random magic numbers in code aren't nice :slightly_smiling_face:

Panaetius (dismissed):

Panaetius (approved):

PR #1518, opened 2023-06-21
feat(app): use PV to hibernate/resume sessions

mohammad-alisafaee:
Adds a `/notebooks/servers/<server_name>` endpoint with `PATCH` method that accepts a `state` argument. Setting `state` to `hibernated` makes the server to go to hibernation and setting `state` to `running` resumes a hibernated server.

When hibernating a server:

- `spec.jupyterServer.hibernated` is set to `True` in the server's manifest file.
- Server's sidecar is queried (via JSON RPC) and some info about the repository (commit, branch, dirty, synched) along with hibernation date are added to server's annotations.
- If querying the sidecar fails, all those annotation except hibernation date will be empty.
- By seeing `spec.jupyterServer.hibernated` set to true, Amalthea scales server's statefulset to have 0 replicas and set's server's `state` to hibernated.

When resuming:

- `spec.jupyterServer.hibernated` is set to `False` in the server's manifest file.
- All server's hibernation annotations are cleared (including the date)
- By seeing `spec.jupyterServer.hibernated` set to false, Amalthea scales server's statefulset to have 1 replica.

It also adds a `renku.io/hibernatedSecondsThreshold` annotation to the server that shows when a hibernated sessions will be deleted.

Related: https://github.com/SwissDataScienceCenter/amalthea/pull/344

/deploy #persist extra-values=notebooks.userSessionPersistentVolumes.enabled=true,notebooks.amalthea.image.repository=renku/amalthea,notebooks.amalthea.image.tag=db70f80ef1983d49a1b167faa57d02d9d8b9554f renku-ui=leafty/2526-feature-persistent-sessions


renku_notebooks/api/amalthea_patches/jupyter_server.py
```diff
@@ -15,7 +15,7 @@ def env(server: "UserServer"):
             "path": "/statefulset/spec/template/spec/containers/0/env/-",
             "value": {
                 "name": "GIT_AUTOSAVE",
-                "value": "1" if server.autosave_allowed else "0",
+                "value": "0",
```
olevski, line 18: This whole patch that adds the GIT_AUTOSAVE environment variable can be removed.

renku_notebooks/api/classes/server.py
```diff
@@ -438,6 +431,12 @@ def get_annotations(self):
             f"{prefix}projectName": self.project,
             f"{prefix}requested-image": self.image,
             f"{prefix}repository": None,
+            f"{prefix}hibernation": None,
+            f"{prefix}hibernation-branch": None,
+            f"{prefix}hibernation-commit-sha": None,
+            f"{prefix}hibernation-dirty": None,
+            f"{prefix}hibernation-synchronized": None,
+            f"{prefix}hibernation-date": None,
```
olevski, line 439: Should we just nest this and store the annotation as json? What do you think? I am going back and forth on the idea.

olevski: So it would be something like:
```
f"{prefix}hibernation": '{"branch": "fdsfsd", "date": "some-date"}'
```

Then when we return this from the notebooks service in the API we use json.load and run it through marshmallow.

renku_notebooks/api/classes/k8s_client.py
```diff
@@ -115,6 +120,119 @@ def create_server(self, manifest: Dict[str, Any]) -> Dict[str, Any]:
         )
         return server
 
+    def hibernate_server(self, server_name: str, access_token: str):
```
olevski, line 123: The k8s client should not worry about calling the git sidecar. So we should put this in a separate util function or a separate adapter/class.

I would even go as far as only adding a "patch" command on the k8s client that accepts the patch and server name and lightly wraps the k8s client command. 

The k8s client should only worry about running the right k8s command in the right namespace.

renku_notebooks/api/classes/k8s_client.py
```diff
@@ -324,6 +443,51 @@ def create_server(self, manifest: Dict[str, Any], safe_username: str):
             return self.renku_ns_client.create_server(manifest)
         return self.session_ns_client.create_server(manifest)
 
+    def hibernate_server(self, server_name: str, access_token: str, safe_username: str):
```
olevski, line 446: Same as above this method and the resume method should go somewhere else and the k8s client should just patch.

renku_notebooks/api/notebooks.py
```diff
@@ -295,11 +295,74 @@ def launch_notebook(
...
     return NotebookResponse().dump(UserServerManifest(manifest)), 201
 
 
+@bp.route("servers/<server_name>", methods=["PATCH"])
+@use_args(
+    {"server_name": fields.Str(required=True)}, location="view_args", as_kwargs=True
+)
+@use_args(PatchServerRequest(), location="json", as_kwargs=True)
+@authenticated
+def hibernate_or_resume_server(user, server_name, state):
```
olevski, line 310:
```suggestion
def patch_server(user, server_name, state):
```

In the future this will grow to do more changes not just hibernate and resume.

renku_notebooks/api/classes/k8s_client.py
olevski: The logic for hibernate and resume should go to `classes/server.py` right next to the start method in the `UserServer` class. I know this class is insane but my plan is to refactor this in the future and slim it down to only do start/stop/delete/hibernate/resume and remove all the other checks and stuff in other adapters/classes.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -299,14 +298,21 @@ tolerations: []
...
   maxAgeSecondsThreshold:
     anonymous: 0 # do not cull sessions solely based on age (use activity)
     registered: 0 # do not cull sessions solely based on age (use activity)
+  # How long before a hibernated session is removed from the cluster.
+  # Setting a value of zero below makes it so that a session is never removed.
+  # Anonymous users don't have hibernated sessions, so, we put the minimum
+  # threshold value for them.
+  hibernatedSecondsThreshold:
+    anonymous: 1 # 1 second
+    registered: 86400 # 24 hours
```
leafty, line 315: This should be at least 72 hours, and most likely close to 7 days or more.

mohammad-alisafaee: This can be set in deployment's values file later if needed. I've changed the default to 72 hours.

renku_notebooks/api/classes/k8s_client.py
mohammad-alisafaee: As discussed, I moved this to a utility file and do it in the `notebooks` handlers. I've renamed the methods in the k8s client to `patch_server` which only call the patching API.

renku_notebooks/api/classes/server.py
mohammad-alisafaee: UI requested to have them as separate fields since it's easier to deal with. We also have a field like that (`"renku.io/hibernation"`) that includes the json for hibernation values.

renku_notebooks/api/classes/k8s_client.py
olevski: Thank you!

renku_notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -31,14 +31,9 @@ def git_clone(server: "UserServer"):
         {"name": "GIT_CLONE_COMMIT_SHA", "value": server.commit_sha},
         {"name": "GIT_CLONE_BRANCH", "value": server.branch},
         {
-            # used only for naming autosave branch
             "name": "GIT_CLONE_USER__USERNAME",
             "value": server._user.username,
```
olevski, line 35: We can remove this since it is only used to name the autosave branch... I think the comment is still true.

renku_notebooks/api/classes/k8s_client.py
```diff
@@ -115,6 +117,22 @@ def create_server(self, manifest: Dict[str, Any]) -> Dict[str, Any]:
...
+                namespace=self.namespace,
+                plural=self.amalthea_plural,
+                name=server_name,
+                body=patch,
+            )
+        except ApiException as e:
+            logging.exception(f"Cannot patch server {server_name} because of {e}")
+            raise PatchServerError()
+
+        return status
```
olevski, line 134: It is fine to call this status... but I am pretty sure it is the manifest of the jupyterserver. And you can return it from this function and use it in other places. Esepecially since it would be nice if the patch method responded with the new server manifest rather than an empty body.

helm-chart/renku-notebooks/values.yaml
```diff
@@ -299,14 +296,21 @@ tolerations: []
...
     registered: 86400 # 24 hours
   maxAgeSecondsThreshold:
     anonymous: 0 # do not cull sessions solely based on age (use activity)
     registered: 0 # do not cull sessions solely based on age (use activity)
+  # How long before a hibernated session is removed from the cluster.
+  # Setting a value of zero below makes it so that a session is never removed.
+  # Anonymous users don't have hibernated sessions, so, we put the minimum
+  # threshold value for them.
+  hibernatedSecondsThreshold:
+    anonymous: 1 # 1 second
```
olevski, line 312: Remove this from here and hardcode a 1 in the config. Note in the comments above that the anonymous sessions cannot be hibernated. Also where you hardcode the 1 in the config mention in a comment why we do this.

helm-chart/renku-notebooks/templates/statefulset.yaml
```diff
@@ -87,9 +87,9 @@ spec:
             - name: NB_SSH_ENABLED
               value: {{ .Values.ssh.enabled | quote }}
             - name: NB_SESSIONS__CULLING__REGISTERED__IDLE_SECONDS
-              value: {{ .Values.culling.idleThresholdSeconds.registered | quote }}
+              value: {{ .Values.culling.idleSecondsThreshold.registered | quote }}
             - name: NB_SESSIONS__CULLING__ANONYMOUS__IDLE_SECONDS
-              value: {{ .Values.culling.idleThresholdSeconds.anonymous | quote }}
+              value: {{ .Values.culling.idleSecondsThreshold.anonymous | quote }}
```
olevski, line 92: You are missing here somewhere the hibernation threshold, right? Same in the test.yaml.

olevski (changes requested):
I got really picky and I apologize. But I think most of it is for weird edge cases that may or may not come up. But better to handle this now.

If you want open an issue for some things so that we can do them later. Or maybe I am just too paraniod.

renku_notebooks/api/schemas/servers_get.py
```diff
@@ -164,7 +158,7 @@ def get_user_correctable_message(exit_code):
                 143: default_server_error_message,
                 200: "Cannot clone repository: Unhandled git error.",
                 201: "Cannot clone repository: Git remote server is unavailable. Try again later.",
-                202: "Cannot clone repository: "
+                202: "Deprecated: Cannot clone repository: "
```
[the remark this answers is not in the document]
mohammad-alisafaee: This is used when autosave branch doesn't have a correct format. Since we don't use them anymore, this shouldn't happen.

renku_notebooks/api/notebooks.py
```diff
@@ -295,11 +301,143 @@ def launch_notebook(
...
+
+    logging.warning(f"State is {state}")
+    logging.warning(
+        f"H = {PatchServerStatusEnum.Hibernated.value}, R = {PatchServerStatusEnum.Running.value}"
+    )
+
+    server = config.k8s.client.get_server(server_name, user.safe_username)
+
+    if state == PatchServerStatusEnum.Hibernated.value:
+        # NOTE: Do nothing if server is already hibernated
```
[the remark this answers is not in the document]
mohammad-alisafaee: The issue with repeated patching is that we can't get repository's status the second time (from the git sidecar). In such cases, we shouldn't overwrite hibernation annotations that's probably retrieved from the sidecar the first time. Therefore, we always need to check if the sessions is hibernated or not.
I can change the code to always patch but don't overwrite hibernation annotations but that's basically the same as not patching.

Resuming is ok and can be repeated with no problem since we remove those annotations.

renku_notebooks/api/notebooks.py
```diff
@@ -295,11 +307,140 @@ def launch_notebook(
...
+        if server and server.get("spec", {}).get("jupyterServer", {}).get(
+            "hibernated", False
+        ):
+            logging.warning(f"Server {server_name} is already hibernated.")
+
+            return NotebookResponse().dump(UserServerManifest(server)), 204
+
+        status = get_status(server_name=server_name, access_token=user.access_token)
```
olevski, line 384: This will return None if the sidecar is not available. We should handle that. The return type of get status is wrong. Sorry we do not have mypy here.

renku_notebooks/api/notebooks.py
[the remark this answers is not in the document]
olevski: So I think it is easier to do the following on patch:
1. Patch the "hibernated" flag in the spec
2. Attempt to get the status from the sidecar:
  - if you cannot get the status leave the hibernation annotations alone
  - if you can get the status then update the annotations

This way you remove one call to the k8s api on patching. And yes if the session is patched already then essentially this does not do anything. Which is exactly right.

olevski (approved):

renku_notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -30,15 +30,6 @@ def git_clone(server: "UserServer"):
         },
         {"name": "GIT_CLONE_COMMIT_SHA", "value": server.commit_sha},
         {"name": "GIT_CLONE_BRANCH", "value": server.branch},
-        {
-            # used only for naming autosave branch
-            "name": "GIT_CLONE_USER__USERNAME",
-            "value": server._user.username,
-        },
```
leafty, line 37: This is still required:
```
Error in sys.excepthook:
  File "/app/.venv/lib/python3.11/site-packages/dataconf/main.py", line 26, in parse
    return utils.__parse(conf, clazz, "", strict, ignore_unexpected)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.11/site-packages/dataconf/utils.py", line 76, in __parse
    fs[f.name] = __parse(
                 ^^^^^^^^
  File "/app/.venv/lib/python3.11/site-packages/dataconf/utils.py", line 85, in __parse
    raise MalformedConfigException(
dataconf.exceptions.MalformedConfigException: expected type <class 'git_services.init.config.User'> at .user, no username found in dataclass
```

renku_notebooks/api/notebooks.py
```diff
@@ -295,11 +307,138 @@ def launch_notebook(
...
+        if status:
+            hibernation = {
+                "branch": status.get("branch", ""),
+                "commit": status.get("commit", ""),
+                "dirty": not status.get("clean", True),
+                "synchronized": status.get("ahead", 0) == status.get("behind", 0) == 0,
+            }
+
+        hibernation["now"] = datetime.now(timezone.utc).isoformat(timespec="seconds")
```
leafty, line 390:
```suggestion
        hibernation["date"] = datetime.now(timezone.utc).isoformat(timespec="seconds")
```

leafty (changes requested):

renku_notebooks/api/notebooks.py
mohammad-alisafaee: Good catch!

olevski (approved):

leafty (approved):

PR #1568, opened 2023-08-08
fix(app): properly recover LFS from autosave

olevski:
This is an issue that Till discovered that if there are LFS files in the autosave branch they do not get recovered. The reason for this is that by default we do not pull LFS files (even when recovering an autosave). So any files that were checked in LFS and created with the autosave are lost and only the LFS pointers (which are not valid anymore) are left. And the pointers point to the non-existent branch and LFS data.

This corrects the problem.

/deploy #persist


mohammad-alisafaee (approved):
LGTM, thank you!

PR #1653, opened 2023-10-05
feat: make an Image class

olevski:
VSCode provides a much nicer diff on the image class. Because I moved the file and then edited it. So the changes are much smaller than what is shown.

Note that the target branch is not master.


Panaetius (approved):
Thank you, looks much nicer now!

PR #1662, opened 2023-10-11
fix: scientific notation integer parsing in config

olevski:
Helm will convert a large integer (specified normally in the values file) like `3000000` to scientific notation like `3e6` in the templates.

And python `int()` crashes when it sees `3e6` and you have to use `float()` to parse it.

Also I have the same function repeated because it is in two separate modules that are at the same level and they share no code. So to not repeat that one simple function I would have to do a lot unnecessary gymnastics just for one small really simple function.

### DIFF
--- a/git_services/git_services/sidecar/config.py
+++ b/git_services/git_services/sidecar/config.py
@@ -1,19 +1,16 @@
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Text, TypeVar, Union
+from typing import Any, Text, TypeVar, Union
 
 import dataconf
 
 from git_services.cli.sentry import SentryConfig
 
 
-def _parse_value_as_numeric(val: Any, parse_to: Callable) -> Union[float, int]:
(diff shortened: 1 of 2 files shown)


git_services/git_services/sidecar/config.py
```diff
@@ -29,8 +26,8 @@ class Config:
     git_proxy_health_port: Union[Text, int] = 8081
 
     def __post_init__(self):
-        self.port = _parse_value_as_numeric(self.port, int)
-        self.git_proxy_health_port = _parse_value_as_numeric(self.git_proxy_health_port, int)
+        self.port = _parse_value_as_int(self.port)
```
Panaetius, line 29: **question (non-blocking):** Do ports actually have this issue, since they only go to 65535?

olevski: Probably not but I am not sure at what value k8s decides to convert large ints to scientific notation

mohammad-alisafaee (approved):

PR #1672, opened 2023-10-18
fix: overwrite existing files with storage mounts

olevski:
There can be weird behaviour if someone overwrites an existing folder with a cloudstorage mount.

So we do not allow this.


olevski:
This works as expected. You can see it deployed here: https://github.com/SwissDataScienceCenter/renku/pull/3214

Panaetius (approved):

PR #1683, opened 2023-10-30
fix: number and type of gunicorn workers

olevski:
So this is something that we have "always done" I think. But according to the flask [docs here](https://flask.palletsprojects.com/en/3.0.x/deploying/gunicorn/#async-with-gevent-or-eventlet), when deploying with gunicorn there is no point in using `gevent` async workers if you do not explicitly use `gevent` in your code. And we do not do this. So we get no benefit.

Also we only used a single worker because we did not set anything and the default is 1.

The [docs](https://docs.gunicorn.org/en/stable/design.html#how-many-workers) for gunicorn recommend 4-12 workers. And because in production we are kind of going for smaller many replicas rather than a large one I went for the smaller number of workers.

/deploy


Panaetius (approved):

rokroskar:
@olevski @Panaetius should this still be merged?

PR #1695, opened 2023-11-02
fix: properly patch tolerations and affinities

olevski:
/deploy renku-data-services=fix-tolerations-affinities


mohammad-alisafaee (approved):
LGTM, thanks!

PR #1700, opened 2023-11-07
fix: do not remove the repo in cloning step

olevski:
This caused sessions that are hibernated to lose work.


mohammad-alisafaee (approved):
LGTM, thanks!

PR #1707, opened 2023-11-15
feat: rclone csi support

Panaetius:
adds initial support to mount sessions using rclone csi
/deploy


renku_notebooks/api/classes/cloud_storage/existing.py
```diff
@@ -1,59 +1,37 @@
...
 
-from ....config import config
-
 
 @dataclass
 class ExistingCloudStorage:
-    bucket: str
-    endpoint: str
+    remote: str
+    type: str
```
olevski, line 8: Not the end of the world, but it feels like it may be a good idea to make this an Enum?

Or will this then have to have every possible value that Rclone supports?

olevski: Also a general question here:
Is this now capturing old cloud storage - backed by datashim as well as new stuff backed by rclone? If not how do we support datashim on sessions that already exist?

renku_notebooks/api/schemas/cloud_storage.py
```diff
@@ -29,80 +30,153 @@ def validate_storage(self, data, **kwargs):
...
+        self, base_name: str, namespace: str, labels={}, annotations={}
+    ) -> List[Dict[str, Any]]:
+        patches = []
+        patches.append(
+            {
+                "type": "application/json-patch+json",
+                "patch": [
+                    {
+                        "op": "add",
+                        "path": f"/{base_name}-pv",
```
olevski, line 86: Reminder to take this out when the CSI driver is ready. Also we are not merging this like this until the driver is ready, correct?

renku_notebooks/api/schemas/cloud_storage.py
```diff
@@ -29,80 +30,153 @@ def validate_storage(self, data, **kwargs):
...
+                        "path": f"/{base_name}-pvc",
+                        "value": {
+                            "apiVersion": "v1",
+                            "kind": "PersistentVolumeClaim",
+                            "metadata": {
+                                "name": base_name,
+                                "namespace": namespace,
+                            },
+                            "spec": {
+                                "storageClassName": "rclone",
```
olevski, line 123: It would be nice if the class name here came from the config.

renku_notebooks/api/schemas/cloud_storage.py
```diff
@@ -20,6 +20,7 @@ class Meta:
     )
     storage_id: Optional[str] = fields.Str(load_default=None, allow_none=True)
     readonly: bool = fields.Bool(load_default=True, allow_none=False)
+    _mount_folder = None
```
olevski, line 23: As far as I can tell this is completely unused, can we remove it?

renku_notebooks/api/schemas/config_server_options.py
```diff
@@ -36,16 +36,10 @@ class Meta:
     lfs_auto_fetch = fields.Bool(required=True)
 
 
-class CloudStorageServerOption(Schema):
-    """Used to indicate in the server_options endpoint which types of cloud storage is enabled."""
+class ServerOptionsEndpointResponse(ServerOptionsChoices):
+    """Used to serialize the server options sent out through the server_options endpoint."""
 
-    s3 = fields.Nested(
+    cloudstorage = fields.Nested(
```
olevski, line 42: This requires a helm chart change. Just a reminder.

renku_notebooks/api/schemas/version.py
```diff
@@ -5,7 +5,7 @@ class NotebooksServiceInfo(Schema):
     """Various notebooks service info."""
 
     anonymousSessionsEnabled = fields.Boolean(required=True)
-    cloudstorageEnabled = fields.Dict(required=True, keys=fields.String, values=fields.Boolean)
+    cloudstorageEnabled = fields.Boolean(required=True)
```
olevski, line 8: Just a reminder - this requires a UI change. I think it is already handled but just in case.

renku_notebooks/api/classes/data_service.py
```diff
@@ -53,26 +62,44 @@ def get_storage_by_id(self, user: User, project_id: int, storage_id: str) -> Clo
...
+    ) -> None:
         res = requests.post(self.storage_url + "/storage_schema/validate", json=configuration)
         if res.status_code == 422:
             raise InvalidCloudStorageConfiguration(
                 message=f"The provided cloud storage configuration isn't valid: {res.json()}",
             )
-        if res.status_code != 204:
+        if res.status_code != 204 and res.status_code != 501:
+            # 501 means connection testing isn't supported for the specified provider but the
+            # schema passed validation.
```
olevski, line 78: I thought that we were going to have a separate endpoint for checking the config (validation) and another to check if a connection with the validated config is possible. From this it does not look like that is the case.

olevski (changes requested):

renku_notebooks/api/classes/cloud_storage/existing.py
Panaetius: since this is only for showing cloudstorage for sessions and the UI doesn't show cloud storage, it's irrelevant. I remember we discussed this.

renku_notebooks/api/schemas/cloud_storage.py
Panaetius: we're not releasing it without that. not sure about merging

renku_notebooks/api/schemas/config_server_options.py
Panaetius: there is a corresponding pr in the main repo

olevski (approved):

olevski (approved):

PR #1804, opened 2024-03-07
renku 2.0 sessions which work

leafty:
Sessions can be started with one repository from GitLab.

See: https://github.com/SwissDataScienceCenter/renku-ui/pull/3053


rokroskar:
@leafty @m-alisafaee what needs to happen with this PR? Is it needed for the current project page implementation?

leafty:
> @leafty @m-alisafaee what needs to happen with this PR? Is it needed for the current project page implementation?

Oh, I think this should be closed now

PR #1847, opened 2024-04-24
feat: add sidecar for mounting secrets

Panaetius:
/deploy renku=release-0.52.x


secrets-mount/secrets_mount/__main__.py
```diff
@@ -0,0 +1,100 @@
...
+    if user_token is None:
+        logging.warn("No user token set, skipping secrets mount")
+        return
+
+    data_svc_url = os.environ.get("DATA_SERVICE_URL")
+    if data_svc_url is None:
+        logging.error("DATA_SERVICE_URL not set")
+
+    secrets_mount_path = Path(os.environ.get("SECRETS_MOUNT_PATH", "/secrets_enc/"))
+    if not secrets_mount_path.exists():
```
sgaist, line 80: I would suggest to be more explicit with the names. since we have encrypted and decrypted secrets that are managed.

```suggestion
    encrypted_secrets_mount_path = Path(os.environ.get("ENCRYPTED_SECRETS_MOUNT_PATH", "/secrets_enc/"))
    if not encrypted_secrets_mount_path.exists():
```

secrets-mount/secrets_mount/__main__.py
```diff
@@ -0,0 +1,100 @@
...
+        logging.info("no encrypted secrets mounted on session, skipping secrets mount")
+        return
+    secrets_target_path = Path(os.environ.get("SECRETS_TARGET_PATH", "/secrets/"))
+    secrets_target_path.mkdir(parents=True, exist_ok=True)
```
sgaist, line 84: Same as above, more explicit names

```suggestion
    decrypted_secrets_mount_path = Path(os.environ.get("DECRYPTED_SECRETS_MOUNT_PATH", "/secrets/"))
    decrypted_secrets_mount_path.mkdir(parents=True, exist_ok=True)
```

secrets-mount/secrets_mount/__main__.py
```diff
@@ -0,0 +1,100 @@
...
+    if user_key is None:
+        return
+
+    logging.info("Decrypting user secrets")
+
+    for child in secrets_mount_path.iterdir():
+        if child.is_dir():
+            continue
+
+        decrypt_secret(child, secrets_target_path, user_key)
```
sgaist, line 97:
```suggestion
        decrypt_secret(child, decrypted_secrets_mount_path, user_key)
```

secrets-mount/secrets_mount/__main__.py
sgaist, line 84: One thing that is not really clear here: why create the folder ?

From a container point of view, the decrypted secrets volume should be mounted at the path indicated in the request and thus its content be a the first level, no ?

sgaist (changes requested):
It's looking good

secrets-mount/secrets_mount/__main__.py
Panaetius: that is true. I only added it out of habit when working with folders in python, but can also remove it.

sgaist (changes requested):
While reviewing something came to my mind: shouldn't this element be part of the secrets storage code base so all the encoding and decoding logic are in the same place ?

Other than that, the implementation looks overall good.

secrets-mount/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    chown -R 1000:1000 /app 
+USER 1000:1000
+WORKDIR /app
+RUN python3 -m pip install --user pipx && \
+    python3 -m pipx ensurepath && \
+    /home/jovyan/.local/bin/pipx install poetry && \
+    python3 -m venv .venv
+COPY poetry.lock pyproject.toml ./
+RUN /home/jovyan/.local/bin/poetry export --only main --without-hashes -o requirements.txt && \
+    .venv/bin/pip install -r requirements.txt --prefer-binary
```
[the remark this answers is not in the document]
Panaetius: I don't think these are equivalent. `poetry install` installs from `pyproject.toml`, whereas `poetry export` exports based on `poetry.lock`. This is here to make builds more reproducible.

Panaetius:
well we keep all the sidecars\init containers in this repository so they're all together. I think the proper solution would be if the code from data services were published as a package that can be imported here, though we don't necessarily want to have a package for data-services public, as that would mean some commitment to keep the python api stable, which we do not want (only the REST api).

I'd leave it as is for now, as this will be a non-issue once we move notebooks into the data-services repo.

secrets-mount/secrets_mount/__main__.py
```diff
@@ -0,0 +1,101 @@
...
+        return
+
+    data_svc_url = os.environ.get("DATA_SERVICE_URL")
+    if data_svc_url is None:
+        logging.error("DATA_SERVICE_URL not set")
+        return
+
+    encrypted_secrets_mount_path = Path(os.environ.get("ENCRYPTED_SECRETS_MOUNT_PATH", "/secrets_enc/"))
+    if not encrypted_secrets_mount_path.exists():
+        logging.info("no encrypted secrets mounted on session, skipping secrets mount")
```
olevski, line 80: If you dont add `logging.basicConfig(level=logging.INFO)` somewhere at the beginning of `main` here then only warning and above will be logged. And all `logging.info` calls will not show up at all.

olevski (changes requested):
Just a few minor questions and potential changes. If you want to ignore them or address later let me know and I will approve.

olevski (approved):

PR #1857, opened 2024-05-02
fix: project path case sensitivity

olevski:
/deploy


olevski:
Different gitlab api endpoints return all-lowercase or case sensitive names for the project name/path. But for mount locations or work-dirs we use the case sensitive version.

For example try to launch a session from this project on dev.renku.ch and the work dir will be wrong.

Panaetius (approved):

leafty:
@olevski For Renku 2.0, all sessions have their working directory set to `<image_work_dir>/work`, so this should not be an issue.

I will check that this case is handled correctly in #1849.

PR #1866, opened 2024-05-10
fix: make dataconf ignore unexpected env vars

Panaetius:
/deploy renku=add-missing-default-value renku-data-services=v0.9.0


olevski (approved):

olevski:
I know this is draft. I will re-approve if needed. Let me know.

olevski:
@Panaetius you should target data service by tag. Because I merged and released the authz PR earlier today. And that actually also requires some changes on the helm chart. I thought we added all of those but we did not. I updated the description.

PR #1873, opened 2024-05-15
feat: restrict secrets mount paths

Panaetius:
limits where users are allowed to mount secrets to improve security. Blacklists "bin", "sbin", "usr", "dev", "proc", "sys", "lib" and "var"


sgaist (approved):
LGTM

PR #1920, opened 2024-06-24
feat: add liveness-detector

SalimKayal:
Adds a component that can be used either as a sidecar for liveness detection by watching workdir files for modifications or as an init container to watch for load average + activity on pty on top of that.

### DIFF
--- a/liveness-detector/Dockerfile
+++ b/liveness-detector/Dockerfile
@@ -1,0 +1,32 @@
+# SPDX-FileCopyrightText: 2024 Idiap Research Institute <contact@idiap.ch>
+# SPDX-FileContributor: Salim Kayal <salim.kayal@idiap.ch>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ghcr.io/prefix-dev/pixi:0.24.2-bullseye-slim AS build
+
+COPY pixi.lock /app/pixi.lock
+COPY pyproject.toml /app/pyproject.toml
+COPY liveness_detector /app/liveness_detector
+WORKDIR /app
+
+RUN pixi install -e default
+
+RUN <<EOF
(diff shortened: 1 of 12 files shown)


liveness-detector/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: 2024 Idiap Research Institute <contact@idiap.ch>
+# SPDX-FileContributor: Salim Kayal <salim.kayal@idiap.ch>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ghcr.io/prefix-dev/pixi:0.24.2-bullseye-slim AS build
```
olevski, line 6: suggestion: Lets use bookworm which is the latest stable version of debian.

```suggestion
FROM ghcr.io/prefix-dev/pixi:0.24.2-bookworm-slim AS build
```

I checked and they have the bookworm image.

liveness-detector/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN pixi install -e default
+
+RUN <<EOF
+echo '#!/bin/bash' > /shell-hook.sh
+echo 'export PATH=/app/.pixi/envs/default/bin:${PATH}' >> /shell-hook.sh
+echo 'exec "$@"' >> /shell-hook.sh
+EOF
+
+FROM docker.io/debian:bullseye-slim AS production
```
olevski, line 21:
```suggestion
FROM docker.io/debian:bookworm-slim AS production
```

olevski (changes requested):
Thanks Salim. LGTM. Just a few minor changes with the docker image debian version.

olevski (approved):

PR #1984, opened 2024-09-18
fix: Mount /dev/shm to the correct container.

aledegano:
Currently container #1 is `oauth-proxy` while JupyterServer, is #0.

Additionally use half of the memory requested by the pod instead of half of the storage.

Since we rely on memory there is no need to tie the increased /dev/shm to requesting storage in the session.

NOTE: This system of patching the manifest is very brittle and the bug being fixed by this commit might appear again as soon as the containers sorting changes. We should think about a different way of interacting with the manifests of the user-sessions.

/deploy renku=leafty/fix-renku-jupyterserver

### DIFF
--- a/renku_notebooks/api/amalthea_patches/general.py
+++ b/renku_notebooks/api/amalthea_patches/general.py
@@ -201,32 +201,33 @@ def oidc_unverified_email(server: "UserServer"):
 
 def dev_shm(server: "UserServer"):
     patches = []
-    if server.server_options.storage:
-        patches.append(
-            {
-                "type": "application/json-patch+json",
-                "patch": [
-                    {
-                        "op": "add",
-                        "path": "/statefulset/spec/template/spec/volumes/-",
-                        "value": {
-                            "name": "shm",
-                            "emptyDir": {
-                                "medium": "Memory",
-                                # NOTE: We are giving /dev/shm up to half of the memory request
-                                "sizeLimit": int(server.server_options.storage / 2),
-                            },
+    patches.append(
+        {
+            "type": "application/json-patch+json",
+            "patch": [
+                {
+                    "op": "add",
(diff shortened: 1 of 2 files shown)


renku_notebooks/api/amalthea_patches/general.py
```diff
@@ -201,32 +201,31 @@ def oidc_unverified_email(server: "UserServer"):
...
+            "patch": [
+                {
+                    "op": "add",
+                    "path": "/statefulset/spec/template/spec/volumes/-",
+                    "value": {
+                        "name": "shm",
+                        "emptyDir": {
+                            "medium": "Memory",
+                            # NOTE: We are giving /dev/shm up to half of the memory request
+                            "sizeLimit": int(server.server_options.memory / 2),
```
olevski, line 216: I am not sure that `server.server_options.memory` is always non null. So do a check and put a default if it is not set. 
```suggestion
                            "sizeLimit": int(server.server_options.memory / 2) if isinstance(server.server_options.memory, int) else "1Gi",
```

olevski (changes requested):
Thanks Alessandro. One suggestion to make sure that things are ok if memory is None.

aledegano:
> Thanks Alessandro. One suggestion to make sure that things are ok if memory is None.

Thanks, I've included your change.
I think we always set the memory for sessions, but it doesn't hurt to check!

olevski:
> NOTE: This system of patching the manifest is very brittle and the bug being fixed by this commit might appear again as soon as the containers sorting changes. We should think about a different way of interacting with the manifests of the user-sessions.

I absolutely agree. The new version of amalthea does not have patches like this at all. So when we retire the old version this code will just be thrown out.

olevski (approved):

PR #1998, opened 2024-12-09
fix: correct implementation for multi-arch images

olevski:
Migration of
https://github.com/SwissDataScienceCenter/renku-data-services/pull/526

This should fix the image tests we see failing in other prs.

/deploy

### DIFF
--- a/renku_notebooks/api/classes/image.py
+++ b/renku_notebooks/api/classes/image.py
@@ -1,9 +1,11 @@
+"""Used to get information about docker images used in jupyter servers."""
+
 import base64
 import re
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
-from typing import Any, Optional, cast
+from typing import Any, Optional, Self, cast
 
 import requests
 from werkzeug.datastructures import WWWAuthenticate
@@ -12,9 +14,16 @@ from ...errors.user import ImageParseError
 
 
 class ManifestTypes(Enum):
+    """The mime types for docker image manifests."""
+
     docker_v2: str = "application/vnd.docker.distribution.manifest.v2+json"
+    docker_v2_list: str = "application/vnd.docker.distribution.manifest.list.v2+json"
     oci_v1_manifest: str = "application/vnd.oci.image.manifest.v1+json"
     oci_v1_index: str = "application/vnd.oci.image.index.v1+json"
+
+
+DEFAULT_PLATFORM_ARCHITECTURE = "amd64"
+DEFAULT_PLATFORM_OS = "linux"
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


renku_notebooks/api/classes/image.py
```diff
@@ -41,16 +50,25 @@ def _get_docker_token(self, image: "Image") -> Optional[str]:
...
         realm = params.pop("realm")
+        if not realm:
+            return None
         headers = {"Accept": "application/json"}
         if self.oauth2_token:
             creds = base64.urlsafe_b64encode(f"oauth2:{self.oauth2_token}".encode()).decode()
             headers["Authorization"] = f"Basic {creds}"
         token_req = requests.get(realm, params=params, headers=headers)
-        return token_req.json().get("token")
+        return str(token_req.json().get("token"))
```
m-alisafaee, line 64: Shouldn't we check if `....get("token")` returns `None` and return `None` if it's the case and convert to string otherwise?

renku_notebooks/api/classes/image.py
```diff
@@ -68,17 +86,43 @@ def get_image_manifest(self, image: "Image") -> Optional[dict[str, Any]]:
...
+                ManifestTypes.docker_v2.value,
+                ManifestTypes.oci_v1_manifest.value,
+            ]:
+                headers["Accept"] = media_type
+            res = requests.get(image_digest_url, headers=headers)
+            if res.status_code != 200:
+                headers["Accept"] = ManifestTypes.oci_v1_manifest.value
+                res = requests.get(image_digest_url, headers=headers)
+            if res.status_code != 200:
+                return None
```
m-alisafaee, line 117: nitpick: This whole `if` block should be inside the previous `if` block, since it's acting on the result of the second `get` request on line 115.

leafty (dismissed):
This looks good, thanks! 👏

m-alisafaee (commented):
Thanks Tasko! Please see my comments.

renku_notebooks/api/classes/image.py
olevski: @m-alisafaee maybe I fail to see it but I think there is no difference between these two code snippets:

```python
            res = requests.get(image_digest_url, headers=headers)
            if res.status_code != 200:
                headers["Accept"] = ManifestTypes.oci_v1_manifest.value
                res = requests.get(image_digest_url, headers=headers)
            if res.status_code != 200:
                return None
```

AND what you propose:

```python
            res = requests.get(image_digest_url, headers=headers)
            if res.status_code != 200:
                headers["Accept"] = ManifestTypes.oci_v1_manifest.value
                res = requests.get(image_digest_url, headers=headers)
                if res.status_code != 200:
                  return None
```

So if the first request is 200 then nothing gets executed from the statements below (in either version). 

If the first request is not 200 then the second request runs for sure. And then the last if statement runs depending on the status code of the second request. The only difference with the suggested edit is that an extra if statement gets tested in the PR version and one less if statement gets tested in the changes you propose (in some cases). So I think it is fine to leave this as is.

olevski: Ok sorry for the long explanation I see you marked it as nitpick.

renku_notebooks/api/classes/image.py
olevski: Good point. That is a bit too brave you are right.

m-alisafaee (approved):

PR #2000, opened 2024-12-22
feat: use different user labels in k8s watcher

olevski:
This is necessary because the new and old servers have different labels that indicate the user ID that the session belongs to. So to stay compatible with older sessions and to enable the switch to the data service for v1 sessions this change is required.


Panaetius (approved):

leafty:
FYI, this breaks v1 sessions.

PR #2005, opened 2025-02-12
fix: mount repos in working dir, not (pvc)mount path

Panaetius:
closes https://github.com/SwissDataScienceCenter/renku-data-services/issues/619
closes https://github.com/SwissDataScienceCenter/renku-data-services/issues/631

tested on https://github.com/SwissDataScienceCenter/renku-data-services/pull/643 PR

### DIFF
--- a/git_services/git_services/init/clone.py
+++ b/git_services/git_services/init/clone.py
@@ -1,4 +1,5 @@
 import sys
+from typing import cast
 
 from git_services.cli.sentry import setup_sentry
 from git_services.init import errors
@@ -15,9 +16,9 @@ if __name__ == "__main__":
     git_cloner = GitCloner(
         repositories=config.repositories,
         git_providers=config.git_providers,
-        workspace_mount_path=config.workspace_mount_path,
+        mount_path=config.mount_path,
         user=config.user,
-        lfs_auto_fetch=config.lfs_auto_fetch,
-        is_git_proxy_enabled=config.is_git_proxy_enabled,
+        lfs_auto_fetch=cast(bool, config.lfs_auto_fetch),
+        is_git_proxy_enabled=cast(bool, config.is_git_proxy_enabled),
     )
     git_cloner.run(storage_mounts=config.storage_mounts)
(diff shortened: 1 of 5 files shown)


git_services/git_services/init/cloner.py
```diff
@@ -68,6 +68,11 @@ def _make_dirname(url: str) -> str:
 
 
 class GitCloner:
+    mount_path: Path
```
olevski, line 71: issue: This is not a dataclass but we are kinda using it like one. The stuff that is already there was hardcoded constant type stuff as class variables. But I think it is confusing to have the rest as class variables - and it may lead to bugs. So lets make it a dataclass.

git_services/git_services/init/clone.py
```diff
@@ -15,9 +16,9 @@
     git_cloner = GitCloner(
         repositories=config.repositories,
         git_providers=config.git_providers,
-        workspace_mount_path=config.workspace_mount_path,
+        mount_path=config.mount_path,
```
olevski, line 19: suggestion: workspace_mount_path was only ever used here, so now that we do not use it here we can remove it from the config.

olevski (changes requested):
Thanks Ralf. Two minor issues.

git_services/git_services/init/cloner.py
```diff
@@ -67,31 +67,18 @@ def _make_dirname(url: str) -> str:
...
-        self.repositories: list[Repository] = [
-            Repository.from_config_repo(r, workspace_mount_path=base_path) for r in repositories
-        ]
-        self.git_providers = {p.id: p for p in git_providers}
-        self.workspace_mount_path = Path(workspace_mount_path)
-        self.user = user
-        self.lfs_auto_fetch = lfs_auto_fetch
-        self.is_git_proxy_enabled = is_git_proxy_enabled
-        self._access_tokens: dict[str, str | None] = dict()
+    _access_tokens: dict[str, str | None] = field(default_factory=dict)
```
olevski, line 81:
```suggestion
    _access_tokens: dict[str, str | None] = field(default_factory=dict, repr=False)
```

olevski (changes requested):
Just one small suggestion to set repr=False on the dictionary with access tokens. All else looks good.

olevski (approved):

PR #2007, opened 2025-02-17
feat: add buildruns and taskruns to the k8s-watcher

leafty:
WIP

PR #2015, opened 2025-03-11
fix: toggle on/off watching resources for image builds (#2014)

olevski:
Fixes an issue from #2007.

When image building is not enabled in the RenkuLab instance, do not cache Shipwright BuildRuns and Tekton TaskRuns and do not provide the corresponding API endpoints.

This was already done in a release branch. But it was never merged into master.

### DIFF
--- a/k8s-watcher/config.go
+++ b/k8s-watcher/config.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


sgaist (approved):

mohammad-alisafaee (approved):
LGTM, Thanks!

